PR #7, opened 2022-08-02
WARP-142-change-did-structure-component-to-public

jkvargas:
**What this PR does** 📖

the idea here is to be able to instantiate DID inside Blink. =)

**Additional comments** 🎤


dariusc93:
Thanks for the PR. You would be able to access `DIDKey` by using `DID::as_ref` or `DID::try_from` (though the former would be a preferable option)

jkvargas:
thanks Darius! :)

PR #82, opened 2022-11-11
feat: Implement message attachment

### LINKED ISSUE #72: Send attachment in a message
Implementation of RayGun doesnt have a option to send a file as an attachment. New implementation will introduce a trait that will handle sending a message with an attachment, with the trait function containing an array of paths that is set for being sent with another parameter used for sending a message with it. Another function will be used to download the file attached to the message, if it is still valid. Upon using that function there will be a check to determine if the reference of that file is valid and if not to return an error, otherwise to begin the attempt of downloading the file to the define path in the function.

In warp-rg-ipfs, if constellation is provided or set, it will upload the file there and obtain the `File` from the current directory of that file name, and other metadata, including reference. If its not provided, there should be a code in the backend to handle storing it into the blockstore with the `File` referencing it and send that info with the message. 

Note:
* If the file already exist in the filesystem (including same matching hash), it will send that file instead of reuploading
  * If file exist but with different hash then we will upload it but with the file being renamed (eg `File`, `File (1)`, etc)
* When fetching over ipfs do require that the other node containing the data to be online unless the CID is pinned to a node in which case there will be an attempt to locate the cid over DHT (eg available by warp-mp-ipfs) unless we are already connected to a node containing it.

dariusc93:
**What this PR does** 📖
Implements message attachments in RayGun

Resolves #72 

**Additional comments** 🎤
For now its recommended not to use very large files until optimizations are done upstream.


InfamousVague:
excite

PR #87, opened 2022-11-17
refactor: Use "mailbox" topic approach for friend requests

### LINKED ISSUE #74: Create a global inbox topic for the user
All the client applications should subscribe to a topic that is specific for the peer itself:
eg. `/peer/<DID>`

Messages that are directed to the given user should be sent over that pubsub topic (eg. friend requests)

iltumio:
**What this PR does** 📖
Uses a "mailbox" topic approach for friend requests

Resolves #74

Friend requests just works as before, but they are not using the FRIENDS_BROADCAST channel

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-mp-ipfs/src/store/friends.rs
+++ b/extensions/warp-mp-ipfs/src/store/friends.rs
@@ -473,14 +473,9 @@ impl<T: IpfsTypes> FriendsStore<T> {
 
                     let bytes = serde_json::to_vec(&payload)?;
 
-                    let topic = get_inbox_topic(recipient.clone());
+                    let topic = get_inbox_topic(&request.to());
 
-                    if self
-                        .ipfs
-                        .pubsub_publish(topic, bytes)
-                        .await
-                        .is_err()
-                    {
+                    if self.ipfs.pubsub_publish(topic, bytes).await.is_err() {
                         //Because we are unable to send a single request for whatever reason, kill the iteration
                         //and proceed to the next item in line
                         break;
--- a/extensions/warp-mp-ipfs/src/store/friends.rs
+++ b/extensions/warp-mp-ipfs/src/store/friends.rs
@@ -228,7 +228,7 @@ impl<T: IpfsTypes> FriendsStore<T> {
                     for friend in friends.iter() {
                         let mut list = store.list_all_raw_request().await.unwrap_or_default();
                         // cleanup outgoing
-                        match list
+                        if let Some(req) = list
                             .iter()
                             .find(|request| {
                                 request.request_type() == InternalRequestType::Outgoing
@@ -237,14 +237,11 @@ impl<T: IpfsTypes> FriendsStore<T> {
                             })
                             .cloned()
                         {
-                            Some(req) => {
-                                list.remove(&req);
-                            }
-                            None => {}
-                        };
+                            list.remove(&req);
+                        }
 
                         // cleanup incoming
-                        match list
+                        if let Some(req) = list
                             .iter()
                             .find(|request| {
                                 request.request_type() == InternalRequestType::Incoming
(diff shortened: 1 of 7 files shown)


extensions/warp-mp-ipfs/src/store/friends.rs
```diff
@@ -385,21 +386,28 @@ impl<T: IpfsTypes> FriendsStore<T> {
...
+            let Queue(peer, data, recipient) = item;
             if let Ok(crate::store::PeerConnectionType::Connected) =
                 connected_to_peer(self.ipfs.clone(), *peer).await
             {
                 let bytes = serde_json::to_vec(&data)?;
 
-                self.ipfs
-                    .pubsub_publish(FRIENDS_BROADCAST.into(), bytes)
-                    .await?;
+                let sender = data.clone().sender().unwrap();
```
dariusc93, line 395:
```suggestion
                let sender = data.clone().sender().ok_or(Error::Other)?;
```

extensions/warp-mp-ipfs/src/store/mod.rs
```diff
@@ -26,6 +26,10 @@ pub const IDENTITY_BROADCAST: &str = "identity/broadcast";
 pub const FRIENDS_BROADCAST: &str = "friends/broadcast";
 pub const SYNC_BROADCAST: &str = "/identity/sync/broadcast";
 
+pub fn get_inbox_topic(did: &DID) -> String {
```
dariusc93, line 29: this probably should go into the friends.rs file

extensions/warp-mp-ipfs/src/store/friends.rs
```diff
@@ -155,7 +156,7 @@ impl<T: IpfsTypes> FriendsStore<T> {
         let store = Self {
             ipfs,
             identity,
-            did_key,
+            did_key: did_key.clone(),
```
dariusc93, line 159: cloning shouldnt be when you can use it from `store.did_key`

extensions/warp-mp-ipfs/src/store/friends.rs
```diff
@@ -385,21 +386,28 @@ impl<T: IpfsTypes> FriendsStore<T> {
...
             {
                 let bytes = serde_json::to_vec(&data)?;
 
-                self.ipfs
-                    .pubsub_publish(FRIENDS_BROADCAST.into(), bytes)
-                    .await?;
+                let sender = data.clone().sender().unwrap();
+
+                let recipients = match data.recipients() {
+                    Some(rec_vec) => {rec_vec.into_iter().filter(move |rec| *rec.fingerprint() != sender.fingerprint() ).collect::<Vec<KeyPair>>()},
```
dariusc93, line 398: we could simply this by doing something like

```rust
data.recipients().map(|list| list.iter().filter(move |rec| *rec.fingerprint() != sender.fingerprint() ).collect::<Vec<_>>()).unwrap_or_default()
```

though i dont see this being used anyplace and probably wouldnt be needed

dariusc93 (commented):
I did leave a couple of comments

extensions/warp-mp-ipfs/src/store/mod.rs
iltumio: ok, we can think of a generalization of that concept in the future

extensions/warp-mp-ipfs/src/store/friends.rs
iltumio: ok, done

dariusc93:
I will be doing some side testing before merging this in to confirm a few things internally.

dariusc93:
Will still continue doing some test on this but will merge this as it is now but may change in the future using the same approach

PR #152, opened 2023-01-23
refactor: Change queue in messaging

### LINKED ISSUE #141: Refactor rg-ipfs queue
The current queue is setup as a map which would check for peers connection status and send messages accordingly if they are online, however this would create a bottleneck with large amounts of messages and events stored in queue and would not be scalable since we check one peer at a time. 

The queue should be refactored to be its own struct that would handle each peer as a entry with a separate task performing checks on each entry. When we wish to add something to queue for that specific peer, we would use a function to the queue that would send the data to the entry, If there one available. If there is no entry for said peer then one would be created. Inside each entry would contain a separate tokio task that would run and would perform a check on the peer connectivity. Once the peer is confirmed to be online, we would deliver all the entries supplied. After everything was delivered, we would signal that the queue entry is empty and the queue, which would also have its own task, would check each entry for its status and after seeing any items marked empty would begin a timer which would mark the removal the entry at the end of the timer. 

Each entry would be saved in its own file that would be load into its own entry points and begin the task after loading it into memory. 

Notes:
- Each entry will be encrypted on file but would be decrypted when loading it prior to the task starting
- Loading each entry should be non-blocking (meaning each entry should be loaded in a separate task)
- If an entry is invalid, corrupted, or damaged, to skip it (and eventually override it)
- When a conversation is deleted and there is an entry referencing a conversation, those items in queue should be removed
- If a message is reference in the entry and it is deleted in the conversation, any entries referencing the message should be remove from queue.
- (Optional) The task in the messaging store should signal the queue to shutdown all entry task. To do this, we could notify, a channel, or an atomicbool to signal each entry task to kill its loop so the task can end. This should only be done if we want a graceful shutdown, though not required as once the data is sent to the queue entry, it should automatically save from within that task. 
- (Optional) Though not implemented yet, have a base so offline services are available, it would utilize that first but if its unable to utilize it for whatever reason to fallback to localize queue. 

Will be blocked for now.

dariusc93:
**What this PR does** 📖
Refactors the queue in messaging to have each message stored and process in its own task.

Resolves #141 

**Additional comments** 🎤


dariusc93:
Closing as once #222 is completed, this will get redone into a behaviour instead.

PR #258, opened 2023-06-02
feat(blink): emit event when a participant is speaking

sdwoodbury:
**What this PR does** 📖
first of all, apologies for branching an unmerged branch and continuing work on it. It made the # of commits look enormous. 

Created the following modules
- loudness: calculates loudness using root mean square. Is optimized to minimize the number of mathematical operations required. 
- speech: uses the output of `loudness` to determine when to emit an event, indicating that someone is speaking. uses the loudness level and the time since the last event was emitted. 

Modified `audio::SinkTrack` and `audio::SourceTrack`
- the `init` function now receives a channel, used to emit `BlinkEvent`s when someone is speaking. 
- also initialized with the peer's id (a `DID`) which is intended to be used when recording audio. I plan to forward the RTP packets to a thread, and the peer id is needed there. 

Misc:
- the audio level is sent via a RTP header extension. I don't have an easy way to expose the `MediaEngine` which is needed to get the identifier correctly. For now there's a hack and a comment which describes how the implementation may be changed in the future. 

**Additional comments** 🎤

### DIFF
--- a/warp/src/blink/mod.rs
+++ b/warp/src/blink/mod.rs
@@ -109,11 +109,14 @@ pub enum BlinkEventKind {
     ParticipantLeft { call_id: Uuid, peer_id: DID },
     /// A participant is speaking
     #[display(fmt = "ParticipantSpeaking")]
-    ParticipantSpeaking { call_id: Uuid, peer_id: DID },
+    ParticipantSpeaking { peer_id: DID },
     // todo: maybe don't use this event and just use a timeout.
     /// A participant stopped speaking
     #[display(fmt = "ParticipantNotSpeaking")]
-    ParticipantNotSpeaking { call_id: Uuid, peer_id: DID },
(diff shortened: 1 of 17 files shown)


warp/src/blink/mod.rs
```diff
@@ -47,6 +47,9 @@ pub trait Blink: Sync + Send + SingleHandle + DynClone {
     /// peers included in the Vec<DID>.
     async fn offer_call(
         &mut self,
+        // May want to associate a call with a RayGun conversation.
+        // This field is used for informational purposes only.
+        conversation_id: Option<Uuid>,
```
dariusc93, line 52: if this is `Option::Some`, you might then want to confirm the conversation exist within raygun via `RayGun::get_conversation`, and probably subscribe to the events from raygun to detect if the conversation been deleted so it would drop/end the call (though this, again, is optional)

dariusc93 (approved):

PR #264, opened 2023-06-16
feat(blink): add echo cancellation

sdwoodbury:
**What this PR does** 📖
- A module was added to store the most recently received audio frame from all peers, and use them to cancel echo from the microphone. 
- The echo canceller had to be passed down from host_media to the source and sink tracks. It had to be added to the SourceTrack and SinkTrack trait too. 
- The only thing i'm unsure of is if the Mutex (from parking_lot) would cause too much delay. `parking_lot` is supposed to be fair so hopefully it's OK. But if it turns out that this is a problem, another option is to use channels to send the incoming audio frames to an echo canceller. 
- The implementation could be better. It would probably be better to ensure that the audio frames used to cancel the echo are older than the input (they have to be played through the speakers before they could produce an echo). But this implementation seems to work so long as the laptop speaker volume is below 50%. 

**Additional comments** 🎤


dariusc93:
> The only thing i'm unsure of is if the Mutex (from parking_lot) would cause too much delay. parking_lot is supposed to be fair so hopefully it's OK. But if it turns out that this is a problem, another option is to use channels to send the incoming audio frames to an echo canceller.

To quickly comment, `Mutex` from parking_lot should be fine to use and shouldnt cause to much delay unless one scope is holding onto the lock for a long period of time. In such cases, the moment you no longer need the lock, you can safely drop it before proceeding to any other logic. If concern, nothing is wrong with utilizing std `Mutex`, as the only reason `parking_lot` is used is due to it not being poisoned if a thread panics while a lock is being held.

extensions/warp-blink-wrtc/Cargo.toml
```diff
@@ -26,6 +26,8 @@ serde_cbor = "0.11.2"
 tokio = { version = "1", features = ["full"] }
 warp = { path = "../../warp" }
 webrtc = "0.6.0"
+# if this fails to build on mac, run this: `brew install autoconf automake libtool pkg-config`
```
dariusc93, line 29:
```suggestion
# if this fails to build on mac, run this: `brew install autoconf automake libtool pkg-config`
```

This is only required to build on mac or is there any dependency that is required so it would run for those who are running the application itself?

extensions/warp-blink-wrtc/src/host_media/audio/loudness.rs
```diff
@@ -31,6 +31,7 @@ impl Calculator {
         f32::sqrt(self.ss * self.normalizer)
     }
 
+    #[allow(dead_code)]
```
dariusc93, line 34: Just a little comment marker to come back to this in the future to determine if this will or will not be used. 

```suggestion
    //TODO: Evaluate code to determine if its needed
    #[allow(dead_code)]
```

tools/audio-codec-cli/src/feedback.rs
```diff
@@ -94,3 +99,199 @@ pub async fn feedback(args: StaticArgs) -> anyhow::Result<()> {
     println!("Done!");
     Ok(())
 }
+
+pub async fn echo(args: StaticArgs) -> anyhow::Result<()> {
+    let host = cpal::default_host();
+    let input_device = host.default_input_device().unwrap();
```
dariusc93, line 105: Is it safe to unwrap here instead of returning an error? Would it be consider a bug if this do panic?

tools/audio-codec-cli/src/feedback.rs
```diff
@@ -94,3 +99,199 @@ pub async fn feedback(args: StaticArgs) -> anyhow::Result<()> {
...
+    let output_device = host.default_output_device().unwrap();
```
dariusc93, line 106: ditto

tools/audio-codec-cli/src/feedback.rs
```diff
@@ -94,3 +99,199 @@ pub async fn feedback(args: StaticArgs) -> anyhow::Result<()> {
...
+            enable_delay_agnostic: true,
+            enable_extended_filter: true,
+        }),
+        enable_high_pass_filter: true,
+        ..Default::default()
+    };
+    processor.set_config(config);
+
+    let host = cpal::default_host();
+    let input_device = host.default_input_device().unwrap();
```
dariusc93, line 196: ditto

tools/audio-codec-cli/src/feedback.rs
```diff
@@ -94,3 +99,199 @@ pub async fn feedback(args: StaticArgs) -> anyhow::Result<()> {
...
+    let output_device = host.default_output_device().unwrap();
```
dariusc93, line 197: ditto

tools/audio-codec-cli/src/feedback.rs
```diff
@@ -94,3 +99,199 @@ pub async fn feedback(args: StaticArgs) -> anyhow::Result<()> {
...
+
+    let processor = Arc::new(Mutex::new(processor));
+    let processor2 = processor.clone();
+
+    let input_data_fn = move |data: &[f32], _: &cpal::InputCallbackInfo| {
+        let mut input_fell_behind = false;
+        for &sample in data {
+            input_frame.push(sample);
+            if input_frame.len() == 480 {
+                let mut ap = match processor.lock() {
```
dariusc93, line 226: Is this lock from std or does it reference a result? If the former, you could access the inner value when it errors and use that too, but that may not be safe to do depending on the actual cause of the error.

dariusc93 (commented):
Looks good so far. Left some comments below

tools/audio-codec-cli/src/feedback.rs
sdwoodbury: considering that this is a test tool, i'm not at all concerned if it crashes. 

Still, it does seem silly to call unwrap() in a function that returns a Result. I'll change it.

tools/audio-codec-cli/src/feedback.rs
sdwoodbury: the lock is from std

sdwoodbury:
I converted this to a draft because I believe the echo cancellation is implemented incorrectly. I believe the code needs to ensure that the `render_frames` used were played around the same time that the microphone input was captured. The current implementation uses the most recently received frame.

extensions/warp-blink-wrtc/Cargo.toml
sdwoodbury: those dependencies are needed for any platform. it seems difficult to build on windows

sdwoodbury:
this currently doesn't build for windows. It can't be merged in yet.

Issue #273, opened 2023-07-06
add capability to record calls to warp-blink-wrtc


sdwoodbury:
we want to record audio/video tracks to an .mp4 file. From my reading I believe the `moov` box should be configured for an empty `mdat` box, and the data be added to the file in fragments. This technique allows writing to the file without having to know about the entire "presentation" beforehand.

dariusc93:
If calls are to be recorded, shouldnt there be some type of indication to alert when somebody in the call is recording? Should users have permission markings of sort to allow calls to be recorded in the first place (eg direct calls both users have to consent, while in group calls, the creator of the group has to allow it or whenever a permission system is implemented to grant/revoke permissions to participants)?

sdwoodbury:
> If calls are to be recorded, shouldnt there be some type of indication to alert when somebody in the call is recording? Should users have permission markings of sort to allow calls to be recorded in the first place (eg direct calls both users have to consent, while in group calls, the creator of the group has to allow it or whenever a permission system is implemented to grant/revoke permissions to participants)?

I like the idea of having both participants consent to recording  for direct calls, and using a permission system for group calls. Such a system could be as simple as: host can record (y/n) and other participants can record (y/n).

PR #278, opened 2023-07-17
feat(audio): add test tool to record fragmented mp4 file

sdwoodbury:
**What this PR does** 📖
This PR uses a fork of the [mp4](https://docs.rs/mp4/0.13.0/mp4/) crate to write a fragmented mp4 file. Eventually this will be used to enable Blink to record audio/video calls. I believe that this is a good stopping point and that it would be most beneficial to next add RTP debugging tools to Blink. Hopefully such  a tool would help me determine how to best ensure the timestamps remain synchronized when recording multiple RTP tracks to an MP4 file. 

**Additional comments** 🎤
related to #273


tools/audio-codec-cli/src/encode.rs
```diff
@@ -14,9 +22,296 @@ use webrtc::{
...
+            if let Some(trun) = moof.trafs[0].trun.as_mut() {
+                trun.data_offset = Some(moof_size as i32 + 8);
+                //println!("trun data offset is: {:?}", trun.data_offset);
+            }
+
+            fragment_start_time += num_samples_in_trun as u64;
+            fragment_sequence_number += 1;
+
+            // this commented out code is helpful for debugging invalid mp4 files.
+            //let mfhd_size = moof.mfhd.box_size();
```
dariusc93, line 238: I wonder if this could be put behind a flag so one could use it at runtime for debugging

dariusc93 (approved):

PR #286, opened 2023-07-22
chore: Use OnceLock<T> instead of Arc<RwLock<T>>

dariusc93:
**What this PR does** 📖
- Replace most `Arc<RwLock<T>>` with `OnceLock<T>` where applicable

N/A

**Additional comments** 🎤
`OnceLock`, a threadsafe of `OnceCell` (which is technically a port from `once_cell` crate), was introduced in 1.70. Utilizing this would ensure that we are only setting it once per initialization as well as lock-free access to the underlining data. 

Note: This is not applicable for every data type so it may be used on specific fields where it is read-only or data that is threadsafe itself.


dariusc93:
Closing this to possibly have the initialization process redone (and to remove some of the complexity)

PR #287, opened 2023-07-31
feat(blink): auto-mute

### LINKED ISSUE #271: improve audio in warp-blink-wrtc
desired features:

- [ ] echo cancellation
- [ ] gain control (to adjust peers' loudness)
- [ ] make bitrate configurable

sdwoodbury:
**What this PR does** 📖
This PR addresses the problem of the speakers feeding into the microphone.
- if Blink detects that you are not speaking, it will not send RTP packets.
- if Blink receives RTP packets from someone else, it will mute you for 100ms.

partially fixes #271 

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-blink-wrtc/src/host_media/audio/automute.rs
+++ b/extensions/warp-blink-wrtc/src/host_media/audio/automute.rs
@@ -2,18 +2,18 @@ use anyhow::{bail, Result};
 use once_cell::sync::Lazy;
 use std::{sync::Arc, time::Duration};
 use tokio::sync::mpsc::{self, error::TryRecvError};
-use warp::sync::{Mutex, RwLock};
+use warp::sync::RwLock;
 
 // tells the automute module how much longer to delay before unmuting
 pub struct AudioMuteChannels {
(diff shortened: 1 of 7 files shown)


sdwoodbury:
@dariusc93 

_1. if they both continue to speak will it be a continuous mute for 100ms?_

Muting happens locally. Whoever spoke first doesn't get muted because theoretically their audio would be received first by the other side, who would then mute themselves.
After the 100ms is up, it's possible that the other side ends up getting muted, depending on whose audio arrives at the other side first. 

_2. How will this impact group audio calls where multiple people may be speaking as the same time?_

Unless they're gaming, I don't see this being an issue. Group calls for most settings still involve only one person speaking at a time. I believe there does need to be a way to disable this feature for certain scenarios. I'm not yet sure how to detect if the audio device is a headset or if the host platform is a laptop.

sdwoodbury:
@dariusc93 I tried using a noise cancellation library and I tried removing the noise myself. I think that this is the best we can do at the moment and that it's needed. Before re-opening this for review I plan to add a setting to disable auto-mute in case someone needs that for their call.

dariusc93 (approved):

Issue #315, opened 2023-09-01
bug(calling): Sheldon's windows PC can't send or receive audio


sdwoodbury:
Not sure exactly when this became an issue because ***someone*** never filed a bug report. 

The on the user's device, the` audio-codec-cli` can successfully record and play audio after setting the number of audio channels to 2. (Note that Uplink shouldn't have a problem with unsupported audio I/O configurations because it performs a check changes the configuration if needed). 

User observed this error message: `failed to send sample: ...` which happens when the audio queue is full for a sink track (which deals with receiving audio from someone else).

The audio source track implementation doesn't log when the queue is full. it's possible that the same thing is happening for the source track. 

 It seems like for the sink track, the CPAL stream isn't playing... `build_output_stream` receives a callback which should poll the queue at the audio sampling rate - 48Khz.

sdwoodbury:
https://docs.rs/cpal/0.15.2/cpal/traits/trait.StreamTrait.html#tymethod.play

PR #316, opened 2023-09-01
fix(audio): make blink audio work on windows

sdwoodbury:
**What this PR does** 📖
- calls [this](https://docs.rs/cpal/0.15.2/cpal/traits/trait.StreamTrait.html#tymethod.play) function. Not all platforms play a CPAL stream automatically. `play()` is needed for these cases. 
- removes a command from `blink-cli` which has no effect
- explicitly sets the number of audio channels when initializing Blink. Before, the host's default configuration was used. 

#315 

**Additional comments** 🎤
- needs to be tested using the `blink-cli` with a windows device. 
- testing changing the output device during a call would also be helpful.


sdwoodbury:
Found another bug

PR #317, opened 2023-09-01
fix(audio): make blink audio work on windows and fix audio codecs

sdwoodbury:
**What this PR does** 📖
- calls [play](https://docs.rs/cpal/0.15.2/cpal/traits/trait.StreamTrait.html#tymethod.play) whenever an audio stream is created. Not all platforms play a CPAL stream automatically. `play()` is needed for these cases. 
- changes how audio data is encoded before  being transported over WebRTC. Blink allows the audio I/O devices to have different formats and encodes them to a common format before sending over WebRTC. But that common format wasn't being used - it just so happened that the `audio_source_codec` and `webrtc_codec` default to being the same. But if someone has a 2 channel audio device then the wrong codec was being used to send audio over WebRTC. 
- when an audio device is selected, the `audio_source_codec` and `audio_sink_codec` are now updated to reflect the supported configuration of the new device. 

- removes a command from `blink-cli` which has no effect
- explicitly sets the number of audio channels when initializing Blink. Before, the host's default configuration was used. 

#315 

**Additional comments** 🎤

### DIFF
--- a/tools/blink-cli/src/main.rs
+++ b/tools/blink-cli/src/main.rs
@@ -129,7 +129,7 @@ async fn handle_command(
             let ids = ids.iter().map(|id| {
                 DID::from_str(id).map_err(|e| format!("error for peer id {}: {}", id, e))
             });
-            let errs = ids.clone().filter_map(|x| x.err()).map(|x| x.to_string());
+            let errs = ids.clone().filter_map(|x| x.err());
             let ids = ids.filter_map(|x| x.ok());
 
             let errs: Vec<String> = errs.collect();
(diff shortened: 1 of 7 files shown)


sdwoodbury:
tested between Linux and Windows. needs to be updated to emit an event when there is network delay/packet loss but that's a separate issue.

tools/blink-cli/src/main.rs
```diff
@@ -96,7 +92,6 @@ enum Repl {
     /// specify which speaker to use for output
     ConnectSpeaker { device_name: String },
     /// set the sampling frequency used to send audio samples over webrtc
```
dariusc93, line 94:
```suggestion
```

dariusc93 (approved):

PR #319, opened 2023-09-05
update(Raygun): Improve raygun location attachments

lgmarchi:
### 1. Let possible to send files from constellation and local disk in same message

**What this PR does** 📖

**Additional comments** 🎤


dariusc93 (commented):
Could you change it so the match condition to use `Result` on line 3289 and here do something like the following for the error branch

```
                            Err(e) => {
                                 let stream = async_stream::stream! {
                                    yield (Progression::ProgressFailed { name: path, last_size: None, error: e }, None);
                                };
                                streams.push(stream.boxed());
                            }
```

This would yield an error if one is unable to obtain the item from constellation

PR #323, opened 2023-09-11
feat(blink): add capability to test audio devices

sdwoodbury:
**What this PR does** 📖
The audio device configuration is now represented as a `AudioDeviceConfig` which abstracts the behavior needed to enumerate speakers/microphones and interact with them. The library user can obtain a `AudioDeviceConfig`  from Blink, view the currently selected speaker/microphone, enumerate speakers/microphones, and select/test combinations of them. The library user can then send the modified `AudioDeviceConfig`  back to Blink, telling it to use the selected speaker/microphone. 

The main point of this is to let the user test their speaker and microphone. The rest is just a refactor. Hopefully it's an improvement but if not I can change the Blink API back to what it was before. 

**Additional comments** 🎤


sdwoodbury:
forgot to emit events during audio device test.

sdwoodbury:
combined this with #326

PR #326, opened 2023-09-15
fix(blink): allow arbitrary audio channels

sdwoodbury:
**What this PR does** 📖
This PR improves the way audio channels are handled. Now for webrtc communication, only one audio channel is used. Blink now allows microphones and speakers with arbitrary number of channels. 

Audio hardware is represented using a new struct instead of an `AudioCodec`, because a speaker/microphone doesn't care about the MIME type - it only deals in raw samples. 

The `AudioCodec` struct no longer has a field for channels because for webrtc communication, only one channel is used. 

This PR also changes the way audio hardware is tested. Blink no longer performs that task, but instead returns a new struct: `AudioDeviceConfig`, which lets the user view, select, and test speaker/microphone combinations. The user can then pass that struct back to blink to apply said configuration. 

There was one change to RayGun because I was getting a compiler error. I believe it has no effect on behavior. 

**Additional comments** 🎤


extensions/warp-blink-wrtc/src/lib.rs
```diff
@@ -345,53 +368,129 @@ impl BlinkImpl {
...
+            let mut devices = match host.input_devices() {
+                Ok(d) => d,
+                Err(e) => {
+                    return Err(warp::error::Error::OtherWithContext(format!(
+                        "could not get input devices: {e}"
+                    )));
+                }
+            };
+            let r = devices.find(|x| {
+                if let Ok(name) = x.name() {
```
dariusc93, line 434: Could do something like 

```rust
x.name().map(|name| name == device_name).unwrap_or_default()
```

dariusc93 (approved):

PR #330, opened 2023-09-29
feat(raygun): use MessageOptions.limit when fetching messages

sdwoodbury:
**What this PR does** 📖
if set, the `limit` will limit the number of messages yielded. This makes it easier to fetch the next X messages when scrolling up/down and makes it possible to fetch X messages before/after a pinned message. 

**Additional comments** 🎤


dariusc93 (commented):
Looks good so far. In `MessageOptions` can we set `range` to `None` if a limit is set and vise versa? This would make sure there isnt any conflicting parts between `limit` and `range`

PR #355, opened 2023-11-03
fix(blink): fix signaling when offering a call

sdwoodbury:
**What this PR does** 📖
when offering a call, sometimes the peers haven't discovered each other. in this scenario, Blink should retry. 
This PR adds a `Nofity` struct and spawns a task to send the `Offer` signal until it goes through or is cancelled via `Notify`. 

**Additional comments** 🎤


sdwoodbury:
decided to refactor blink first.

PR #358, opened 2023-11-09
fix(blink): resend signals to ensure all peers connect

sdwoodbury:
**What this PR does** 📖
- refactors blink  (simple_webrtc and host_media were mostly left alone). Most of the logic is now in `blink_controller.rs` instead of split between the blink implementation, `webrtc_controller.rs` (removed) and `call_initiation.rs` (also removed). 

### the new design made it easy to add the following improvements:
- when offering a call, the offer signal is resent upon failure. 
- during a call, all participants announce their presence and deterministically initiate webrtc connections by comparing DIDs. 

### other improvements
- participants now tell the others when they are recording
- someone should be able to be offline when a call is offered, turn on the app, and join the call
- if `BlinkImpl` is cloned and that clone is dropped, it won't shut down the background tasks anymore. 
- someone should be able to join a call and know who is muted/unmuted/recording. This is done by sending the `ParticipantState` in the `Announce` event and removing the other signals. Blink now has an event for `ParticipantStateChanged`. 

Not an issue in Warp but should address the issue where, especially during a group call, not all the participants can hear each other. 

**Additional comments** 🎤


dariusc93 (approved):

extensions/warp-blink-wrtc/src/blink_impl/data/notify_wrapper.rs
```diff
@@ -0,0 +1,12 @@
+use std::sync::Arc;
+use tokio::sync::Notify;
+
+pub struct NotifyWrapper {
+    pub notify: Arc<Notify>,
+}
+
+impl Drop for NotifyWrapper {
+    fn drop(&mut self) {
+        self.notify.notify_waiters();
```
dariusc93, line 10: you should probably check the strong count of the reference so if its the only thing left, it can signal its waiters. Eg

```rust
if Arc::strong_count(&self.notify) == 1 {
   self.notify.notify_waiters();
}
```

sdwoodbury: `notify` gets cloned and passed to some tasks. I think that dropping NotifyWrapper` wouldn't necessarily mean the strong count would be one. I hoped that by putting NotifyWrapper in an Arc, Drop wouldn't be called until all the `Arc<NotifyWrapper>`s were dropped.

PR #376, opened 2023-11-30
feat(blink): refactor blink and improve call quality

sdwoodbury:
**What this PR does** 📖

#### eliminates the problem where audio cuts out by doing the following
- increasing the size of the `HeapRb` used to pass data to/from CPAL
- increase the delay used by the echo canceller
- removed the "optimization" which didn't send opus packets which were mostly silent

#### created a submodule to aid in testing
- `blink-repl` can use `warp-blink-wrtc` with the `loopback` feature to make it send any received audio back to all connected peers. This program can be run on a cloud device, allowing one person to test a call between two separate locations. 

#### refactored warp-blink-wrtc
- Improved the `host_media` module. This was the next logical step after a [previous PR](https://github.com/Satellite-im/Warp/pull/358) improved IPFS signaling. The abstraction over the codec type was removed because it seems that we will always just use opus. The audio sink was split into a decoder task and a receiver task. The audio source was split into an encoder task and a sender task. Opus encoding/decoding is done on dedicated threads. The decoder task decodes in parallel using rayon. 

#### misc fixes
- updated the `AudioDeviceConfig` trait to use a `oneshot::channel` to signal when `test_speaker` and `test_microphone` are done. `Cpal` streams can't be sent safely between threads so these functions, which create said streams, can't use async. 
- fixed a bug in `test_speaker` where the sound continues to play after the test if the program doesn't drop the `AudioDeviceConfigImpl`  struct. 

**Additional comments** 🎤
The stuff about datachannels isn't used yet and was left in for later. Eventually i'd like to use them to determine round trip time between peers, for detecting poor network conditions. That's what I was thinking of doing when I set out on this refactor.


extensions/warp-blink-wrtc/src/simple_webrtc/mod.rs
```diff
@@ -479,6 +481,53 @@ impl Controller {
...
+                    if let Err(e) = tx2.send(EmittedEvents::DataChannelClosed {
+                        peer: dest2.clone(),
+                    }) {
+                        log::error!(
+                            "failed to send data channel closed event for peer {}: {}",
+                            &dest2,
+                            e
+                        );
+                    }
+                    Box::pin(async {})
```
dariusc93, line 500: does the whole return a future? If so maybe we could use `futures::future::ready` instead to wrap the whole scope in it just to make it more idiomatic.

dariusc93 (approved):

PR #412, opened 2024-01-03
fix(messenger): List all commands in help message

zeenix:
**What this PR does** 📖

List all commands in help message. Also make it so that future commands are always added with their help reference included.

### DIFF
--- a/extensions/warp-ipfs/examples/messenger.rs
+++ b/extensions/warp-ipfs/examples/messenger.rs
@@ -190,20 +190,10 @@ async fn main() -> anyhow::Result<()> {
             - If a new conversation is created, the client will automatically switch to it
             - When setting a path to save data, it will resume the last conversation upon loading
             - If you have multiple conversation, use `/set-conversation <id>` to switch conversation
-
-        list of commands:
-            /create <did> - create conversation with another user
-            /remove-conversation - delete current conversation. This will delete it on both ends
-            /set-conversation <id> - switch to a conversation
-            /list-conversations - list all active conversations
-            /list - list all messages in the conversation. This provides more info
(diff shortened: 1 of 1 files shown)


extensions/warp-ipfs/examples/messenger.rs
```diff
@@ -1026,6 +1016,66 @@ enum Command {
...
+        Command::Unpin(_) => "/unpin <all | message-id> - unpin a message in the conversation.",
+        Command::CountMessages => "/count-messages - count messages in the conversation",
+    };
+        help.push_str(cmd_help);
+        help.push('\n');
+    }
+
+    help
+}
+
```
dariusc93, line 1078: I wonder if we can have `Command` implement `Display` in some way (maybe via `derive_more` crate) and then create the results by converting `cmd` to a string. That way, we can have the command have an associated help for it. 

We could probably even go as far as making a macro that would build this out so we can enforce all the commands apart of `Command`.

zeenix: > I wonder if we can have `Command` implement `Display` in some way (maybe via `derive_more` crate) and then create the results by converting `cmd` to a string.

I thought of that but 2 things:

1. We want to display help for all commands, so we'll need to loop through anyway and it won't save us much on LOC (if at all).
2. The strings here are not really string form of the command variants so it'll be strange to use Display/ToString impl for this IMO.

> We could probably even go as far as making a macro that would build this out so we can enforce all the commands apart of `Command`.

Sure, would be great to automate this but maybe another day. I want to get to function changes fast. I've spent enough time cleaning stuff. :laughing:

dariusc93 (approved):

PR #416, opened 2024-01-08
feat(groups): Ability to allow anyone to add new members to a group

### LINKED ISSUE #386: raygun: Support members adding other members to a group
The current group implementation only allows the creator of the group to add/remove members, but the owner may want to allow anyone to add members. In this case, since there is no permissions in place for groups (which will be separate from this issue), we could split the group multiple types:

1. `Group` - For now, will be the same as `ClosedGroup` but may change in the future to recipient a larger scale group (eg community style groups)
2. `ClosedGroup` - Only the owner/creator of the group will be able to add and remove members (default)
3. `OpenGroup` - Allow the members of the group to add others to the group, with the owner being able to remove them.

We can add a new trait function either called:
- `RayGun::set_group_type` where it would accept the conversation id and an enum called `GroupType` which would be as follows
```rust
pub enum GroupType {
    Open,
    Closed
}
```

- `RayGun::set_conversation_type` which would accept the conversation id and allow setting `ConversationType` with `OpenGroup` and `ClosedGroup` being apart of `ConversationType` to be used in the function. 

We can also modify `RayGun::create_group_conversation` to accept a option to create a open or closed group. 

Note:
- If a conversation is already set as group and toggled to be `OpenGroup`, the creator can only then switch it to `ClosedGroup`.
- Will require ignoring recipient list when signing/verifying the conversation document if the group is opened since the current implementation use the list when signing/verifying (to be sure that the list cannot be tampered unless approved by the creator of the conversation)
- The conversation document should include a `exclude` or `block` list by the creator to prevent adding members who the creator of the group has blocked. (Should only be effective for `OpenGroup` since `Group`/`ClosedGroup` should automatically remove them if they are blocked)
- The former function would be preferred to prevent confusion and possible conflicts in the future.

zeenix:
**What this PR does** 📖

This adds ability for non-owners to add new members to a group, if owner choose so at creation time. Removal is still only allowed for owner.

The `messenger` cli's `create-group`command is also updated to set this option.

Fixes #386.

### DIFF
--- a/warp/src/raygun/mod.rs
+++ b/warp/src/raygun/mod.rs
@@ -386,7 +386,7 @@ pub enum ConversationType {
     #[display(fmt = "direct")]
     Direct,
     #[display(fmt = "group")]
-    Group { open: bool },
+    Group { settings: GroupSettings },
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, warp_derive::FFIVec, FFIFree)]
@@ -490,6 +490,31 @@ impl Conversation {
 
     pub fn set_recipients(&mut self, recipients: Vec<DID>) {
         self.recipients = recipients;
+    }
+}
+
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[repr(C)]
+pub struct GroupSettings {
+    // Everyone can add participants, if set to `true``.
+    members_can_add_participants: bool,
+}
+
+impl GroupSettings {
+    pub fn members_can_add_participants(&self) -> bool {
+        self.members_can_add_participants
+    }
+
+    pub fn set_members_can_add_participants(&mut self, val: bool) {
+        self.members_can_add_participants = val;
+    }
+}
+
+impl Default for GroupSettings {
+    fn default() -> Self {
+        Self {
+            members_can_add_participants: false,
+        }
     }
 }
 
@@ -938,7 +963,7 @@ pub trait RayGun:
         &mut self,
         _: Option<String>,
         _: Vec<DID>,
-        _: bool,
+        _: GroupSettings,
     ) -> Result<Conversation, Error> {
         Err(Error::Unimplemented)
     }
--- a/warp/src/raygun/mod.rs
+++ b/warp/src/raygun/mod.rs
@@ -515,17 +515,11 @@ pub enum ConversationSettings {
 
 /// Settings for a direct conversation.
 // Any future direct conversation settings go here.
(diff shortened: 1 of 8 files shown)


warp/src/raygun/mod.rs
```diff
@@ -386,7 +386,7 @@ pub enum ConversationType {
     #[display(fmt = "direct")]
     Direct,
     #[display(fmt = "group")]
-    Group { open: bool },
+    Group { settings: GroupSettings },
```
dariusc93, line 389: If we are to go with this "setting" then it might be best to leave this as `Group` since the conversation type does not intend to store any additional metadata like `GroupSettings` and instead just have this apart of the `ConversationDocument` as a separate field and have a new trait function to obtain the struct from the document.

zeenix: > since the conversation type does not intend to store any additional metadata like `GroupSettings`

That might not be the original intention but since it's group-specific, it's best be only available/accessible through this enum (if kept in `Conversation`, read below).

> instead just have this apart of the `ConversationDocument` as a separate field and have a new trait function to obtain the struct from the document.

So only `ConversationDocument`and not `Conversation`? 🤔

zeenix: We could have another type:

```rust
enum ConversationSettings {
   Direct(DirectionConversationSettings),
   Group(GroupSettings),
}
```

And use that as the direct field but that's really just duplicating `ConversationType` enum.

dariusc93: > That might not be the original intention but since it's group-specific, it's best be only available/accessible through this enum (if kept in `Conversation`, read below).

I dont mind it being apart of a separate function to obtain the `GroupSettings` since storing it through this enum would not really make sense, especially if `GroupSettings` is ever expanded. 

> So only `ConversationDocument`and not `Conversation`? 🤔

Correct, since `ConversationDocument` that is internal, that could be expanded more if needed.

zeenix: I made the changes as we agreed and for now I added settings to `Conversation`as well (because of the conversion to `ConversationDocument` but we can change that later.

However, the group tests now fail because somehow the conversation ends up with `ConversationSettings::Direct`:

```rust
thread 'test::create_empty_group_conversation' panicked at extensions/warp-ipfs/tests/group.rs:42:9:
assertion `left == right` failed
  left: Direct(DirectConversationSettings)
 right: Group(GroupSettings { members_can_add_participants: false })
```

Need to figure out what I missed.

zeenix:
@dariusc93 good to go?

PR #423, opened 2024-01-10
fix: Fix racy cleanups

zeenix:
**What this PR does** 📖

This PR mainly removes racy cleanup code that relies on the value behind `Arc::strong_count` to remain unchanged in the `Drop::drop` impls, which is not a good idea. The docs warn about this:

> Another thread can change the strong count at any time, including
> potentially between calling this method and acting on the result.

In our case, two threads can easily end up dropping the type in question and both `drop` runs thinking that ref count is still 2 and not doing the cleanups.

This PR also:

* Drop manual `Clone` impl of `Tesseract` struct w/ a `derive`.
* Renames `start` methods of task types to `run` as these methods are the task code itself and not just "starting" the task.

### DIFF
--- a/extensions/warp-ipfs/src/store/document/cache.rs
+++ b/extensions/warp-ipfs/src/store/document/cache.rs
@@ -10,6 +10,8 @@ use futures::{
 };
 use libipld::Cid;
 use rust_ipfs::{Ipfs, IpfsPath};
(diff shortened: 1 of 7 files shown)


dariusc93 (approved):
Looks good to me.

PR #424, opened 2024-01-11
chore: Update tesseract test

dariusc93:
**What this PR does** 📖
- Update tesseract test to make sure it locks after the final drop

### DIFF
--- a/warp/src/tesseract/mod.rs
+++ b/warp/src/tesseract/mod.rs
@@ -737,7 +737,6 @@ impl TesseractInner {
         self.soft_unlock.store(false, Ordering::Relaxed);
 
         let _ = self.event_tx.try_broadcast(TesseractEvent::Locked);
-  
     }
 
     fn subscribe(&self) -> BoxStream<'static, TesseractEvent> {
--- a/warp/src/tesseract/mod.rs
+++ b/warp/src/tesseract/mod.rs
@@ -37,7 +37,8 @@ pub enum TesseractEvent {
 
 impl Default for Tesseract {
     fn default() -> Self {
-        let (event_tx, event_rx) = async_broadcast::broadcast(1024);
+        let (mut event_tx, event_rx) = async_broadcast::broadcast(1);
+        event_tx.set_overflow(true);
         Tesseract {
             inner: Arc::new(TesseractInner {
                 internal: Default::default(),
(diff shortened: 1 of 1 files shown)


warp/src/tesseract/mod.rs
```diff
@@ -732,7 +734,9 @@ impl TesseractInner {
         self.enc_pass.write().zeroize();
         self.unlock.store(false, Ordering::Relaxed);
         self.soft_unlock.store(false, Ordering::Relaxed);
-        let _ = self.event_tx.try_broadcast(TesseractEvent::Locked);
+        if self.event_tx.receiver_count() > 1 {
```
zeenix, line 737: This is also racy. You can't rely on numbers to remain the same after the call in a multi-threaded runtime env. In this specific case, this is not needed anyway. `try_broadcast` call will just fail and we ignore the error so nothing really happens.

zeenix (changes requested):
LGTM otherwise.

warp/src/tesseract/mod.rs
zeenix: to explain a bit better the race here: After the receiver count is reported to be 1, another receiver can be initialised from another thread before you act on this outdated information.

dariusc93: > This is also racy. You can't rely on numbers to remain the same after the call in a multi-threaded runtime env. In this specific case, this is not needed anyway. `try_broadcast` call will just fail and we ignore the error so nothing really happens.

Actually, given that we are holding onto the receiver, `try_broadcast` would not fail unless the channel is full or closed and due to the buffer being more than 1 (in this case, its at 1024), its unlikely to fail. Though i dont see any case where this would be racy, the appropriate change would likely to decrease the channel buffer to 1 and enable overflow (so it would pop the last event than just leaving it queued)

zeenix: > the appropriate change would likely to decrease the channel buffer to 1

Just in case you didn't know already, the `receiver_count` is not the amount of buffered messages, it's the number of current receiver instances for this channel.

zeenix (approved):

warp/src/tesseract/mod.rs
dariusc93: > Just in case you didn't know already, the `receiver_count` is not the amount of buffered messages, it's the number of current receiver instances for this channel.

Im aware. The idea was to only broadcast the event if there is more than one receiver (since we hold onto a receiver internally), hence the check. This wouldve prevent queuing the messages since originally the buffer was set to 1024 until `Tesseract::subscribe` is called and the stream is being polled. After your comment and looking over it though, it would make more sense to lower the buffer to 1 and enable the overflow than to check the receiver so atleast the last event is queued (which in this case if one subscribed but havent polled the stream and tesseract drops completely, it would have the lock event there and not any previous event, including any possibly future events)

PR #425, opened 2024-01-12
feat: Add additional reference and an internal document

dariusc93:
**What this PR does** 📖
- Add a thumbnail reference to allow storing thumbnail out of index
- Adds internal documents referencing towards `File`, `Item` and `Directory`
- Add placeholder for later refactoring for attachments. 

**Additional comments** 🎤
- Placeholder will be apart of a refactoring done to have `MessageDocument` store linked attachments that will be resolvable to `File` when resolving the document into `Message`. This would allow us to send over the `MessageDocument` to the topic instead of `Message`, which would reduce the data size being transmitted, remove some limitations (eg file attachments in a single message including thumbnails, though this limit would not be removed but instead would be increased.)
- New documents will be used in a later PR to build graph of the index to be stored in `RootDocument`


extensions/warp-ipfs/src/store/document.rs
```diff
@@ -279,3 +284,93 @@ impl RootDocument {
...
+    pub name: String,
+    pub size: usize,
+    pub thumbnail: Cid,
+    pub file_type: FileType,
+    pub data: Cid,
+}
+
+impl FileAttachmentDocument {
+    pub async fn resolve_to_file(&self, ipfs: &Ipfs) -> Result<File, Error> {
+        let file = File::new(&self.name);
```
zeenix, line 299: is it a name or path? If former, why do we save in $PWD? 🤔

extensions/warp-ipfs/src/store/document.rs
```diff
@@ -279,3 +284,93 @@ impl RootDocument {
...
+        Ok(file)
+    }
+
+    pub fn download<'a, P: AsRef<Path>>(
+        &'a self,
+        ipfs: &'a Ipfs,
+        path: P,
+        members: &'a [PeerId],
+        timeout: Option<Duration>,
+    ) -> BoxStream<'a, Progression> {
```
zeenix, line 334: I think you don't need explicit lifetimes here:

```suggestion
    pub fn download<P: AsRef<Path>>(
        &self,
        ipfs: &Ipfs,
        path: P,
        members: &[PeerId],
        timeout: Option<Duration>,
    ) -> BoxStream<'_, Progression> {
```

extensions/warp-ipfs/src/store/document/files.rs
```diff
@@ -0,0 +1,745 @@
...
+            ipfs.clone(),
+            &Config::development(),
+            event.clone(),
+            Span::current(),
+        )
+        .await?;
+        Ok(store)
+    }
+
+    //TODO: Write test with a larger index and build and resolve graph
```
zeenix, line 297: huge nitpick: please add a space after `//` :)

extensions/warp-ipfs/src/thumbnail.rs
```diff
@@ -44,13 +46,23 @@ impl Default for ThumbnailId {
     }
 }
 
-#[derive(Default, Clone)]
+#[derive(Clone)]
 #[allow(clippy::type_complexity)]
 pub struct ThumbnailGenerator {
-    task: Arc<Mutex<BTreeMap<ThumbnailId, JoinHandle<Result<(ExtensionType, Vec<u8>), Error>>>>>,
+    ipfs: Ipfs,
+    task: Arc<
```
zeenix, line 53: the name is misleading IMO. Should be either `tasks` (plural) or `task_handles`.

extensions/warp-ipfs/src/thumbnail.rs
```diff
@@ -44,13 +46,23 @@ impl Default for ThumbnailId {
     }
 }
 
-#[derive(Default, Clone)]
+#[derive(Clone)]
 #[allow(clippy::type_complexity)]
```
zeenix, line 50: Not related to your PR but IMO it's best to avoid ignoring clippy lints as much as possible. This one for example, is just asking for a simple type alias which would make the code easier to read/follow and so its good advice.

warp/src/constellation/item.rs
```diff
@@ -30,6 +30,15 @@ pub enum FormatType {
     Mime(mediatype::MediaTypeBuf),
 }
 
+impl From<FormatType> for FileType {
```
zeenix, line 33: These types seem identical, why not just have one?

zeenix (changes requested):
I feel I'm not familiar enough with the code to be able to do a thorough enough review but I just looked it from an outsider perspective. LGTM other than the comments.

extensions/warp-ipfs/src/store/document.rs
dariusc93: This is the name of a file, that is recorded apart of [File](https://github.com/Satellite-im/Warp/blob/main/warp/src/constellation/file.rs) for constellation. Not related to the os filesystem itself

extensions/warp-ipfs/src/store/document.rs
dariusc93: Tried it before but compiler calls for lifetime unless im overlooking something here

extensions/warp-ipfs/src/store/document.rs
dariusc93: (which is apart of the internal index)

warp/src/constellation/item.rs
dariusc93: They are identical at the current time. `FormatType` will have additional fields later that would be irrelevant to `FileType`, although previously it did have additional fields that were removed.

extensions/warp-ipfs/src/thumbnail.rs
dariusc93: It is good advice. This was more of me ignoring the lint and forgot about it since I didnt view the type as hard to read, but I did update it in this PR just as a small cleanup.

extensions/warp-ipfs/src/store/document.rs
dariusc93: Since this is commented out for now, this could probably be looked into at a later PR when the changes are implemented since I may need to make changes upstream too.

zeenix: > Tried it before but compiler calls for lifetime unless im overlooking something here

What's the error?

zeenix (changes requested):
Thanks for taking up my advise to keep unused code commented out so compiler doesn't have to work each time for unused code. However, could you please use `/* */` for commenting-out large pieces of code? AFAIK that's the convention and code is more readable that way.

PR #427, opened 2024-01-15
chore: Remove unused ffi code

dariusc93:
**What this PR does** 📖
- Remove unused ffi code, moving code into `deprecated` as a reference for later, separate rewrite of ffi bindings
- Change stream wrappers into alias types
 
- No major changes in the codebase besides removing of unused code that was previously used for ffi

**Additional comments** 🎤


deprecated/warp/src/constellation.rs
```diff
@@ -0,0 +1,393 @@
+#[cfg(not(target_arch = "wasm32"))]
+pub mod ffi {
```
zeenix, line 2: `deprecation` implies the API still part of the lib and available, just marked as "soon to be removed". AFAICT, you're having this code be completely ignored by the compiler.

With git remembering everything (unless we rewrite history on purpose and still `git reflog` remembers locally), I really don't see any value in keeping dead code around. It only confuses people new to the code. If the dead code is needed in the future, it can very easily be revived from git history.

So I strongly recommend we just delete this all.

zeenix (changes requested):

extensions/warp-fs-memory/src/native.rs
```diff
@@ -70,15 +70,13 @@ impl Constellation for MemorySystem {
...
-        let stream = ConstellationProgressStream(
-            futures::stream::once(async move {
-                Progression::ProgressComplete {
-                    name,
-                    total: Some(bytes as _),
-                }
-            })
-            .boxed(),
-        );
+        let stream = futures::stream::once(async move {
```
zeenix, line 73: if the stream always returns only 1 item, I'd not return a stream but a future. The client code can easily convert a stream to a future if it needs to.

deprecated/warp/src/constellation.rs
dariusc93: While internally we could just look at the history (unlikely for one to to rewrite it anyway), I dont see any reason to delete it all right at this moment since it can still be easily be referenced (although `warp-bin` technically outlived its stay there), and given the location of the code its unlikely for a newcomer to get confused per se (since many people have ignored that directory when looking at the code for the first time), however I may rename the `deprecated` directory.

extensions/warp-fs-memory/src/native.rs
dariusc93: I do not follow. Although `warp-fs-memory` is due for a rewrite (its left there mostly for testing when the components were more interchangeable and used for testing purpose), returning a future instead of a stream would break the overall api. Using a stream, even if it is only one item in the stream, would keep it compatible with the api.

deprecated/warp/src/constellation.rs
dariusc93: I went on and removed it and will add a issue to reference back to the commit with them later when we do need to add FFI code again.

PR #428, opened 2024-01-16
chore: Downgrade log level and log specific crates

dariusc93:
**What this PR does** 📖
- Logs specific crates within CI
- Downgrade log level to error in CI to only log errors to prevent the job logs from being flooded

**Additional comments** 🎤


.github/workflows/cargo-test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     env:
       RUSTFLAGS: -D warnings
       RUST_BACKTRACE: full
-      RUST_LOG: debug
+      RUST_LOG: warp_ipfs=error,warp_blink_wrtc=error
```
zeenix, line 34: erorrs are logged by default (as well as warnings) so this change essntially sets the log level to the default for **all** crates.

zeenix:
> * Downgrade log level to error in CI to only log errors to prevent the job logs from being flooded

If the debug log is too verbose, that likely means that we're logging too much to debug, where most of it should go to trace. I'd instead of switch most of `debug!` calls to `trace!`.

Also, there is a [github actions option to explicitly enable logging](https://github.blog/changelog/2022-05-24-github-actions-re-run-jobs-with-debug-logging/). We could look into only setting the log level to debug/trace when this is enabled on a re-run of a failing pipeline.

.github/workflows/cargo-test.yml
dariusc93: To my knowledge, warnings arent logged by default, only errors (eg i usually see errors being logged when running test unless i explicitly set the variable to use a different level), however leaving it set to `RUST_LOG=debug` would log all underlining crates, which imo is not desirable to do on CI due to how much is outputted by all crates providing any logging.

dariusc93:
> > * Downgrade log level to error in CI to only log errors to prevent the job logs from being flooded
> 
> If the debug log is too verbose, that likely means that we're logging too much to debug, where most of it should go to trace. I'd instead of switch most of `debug!` calls to `trace!`.

Some could be switched to a trace level, but some others are fine as is since they themselves arent too verbose themselves.

> Also, there is a [github actions option to explicitly enable logging](https://github.blog/changelog/2022-05-24-github-actions-re-run-jobs-with-debug-logging/). We could look into only setting the log level to debug/trace when this is enabled on a re-run of a failing pipeline.

Hmm yea we could possibly look into that.

.github/workflows/cargo-test.yml
zeenix: > To my knowledge, warnings arent logged by default, only errors (eg i usually see errors being logged when running test unless i explicitly set the variable to use a different level),

That's very strange. I always found it to be the other way around and that would make since warning is between info and error AFAIK. Maybe we're setting some default level in the code (implicitly or explicitly)? 🤔 

> however leaving it set to `RUST_LOG=debug` would log all underlining crates, which imo is not desirable to do on CI due to how much is outputted by all crates providing any logging.

Ah for sure. I agree with not logging all deps at debug level. However, our own crates should be set to debug at least IMO.

PR #430, opened 2024-01-16
feat(conversation): Allow owners to change conversation settings

### LINKED ISSUE #417: raygun: Allow changing closed group chat to open & vice versa
This is the logical follow-up of #386. We should allow owner to open a closed group and the other way around. See description of #386 for details.

zeenix:
**What this PR does** 📖

Currently, this only means changing of "members can add participants" setting of groups conversation. We only allow owners to change the settings.

This also adds an command in `messenger` example CLI app to turn a group open or closed.

* Fixes #417.

### DIFF
--- a/extensions/warp-ipfs/src/store/message.rs
+++ b/extensions/warp-ipfs/src/store/message.rs
@@ -1330,7 +1330,6 @@ impl MessageStore {
                         //      but for now, we can leave this as a silent update since the block list would be for internal handling for now
                     }
                     ConversationUpdateKind::ChangeSettings { settings } => {
-                        conversation.settings = settings;
                         conversation.excluded = document.excluded;
                         conversation.messages = document.messages;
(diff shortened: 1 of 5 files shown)


extensions/warp-ipfs/src/store/message.rs
```diff
@@ -1329,6 +1329,19 @@ impl MessageStore {
                         //TODO: Maybe add a api event to emit for when blocked users are added/removed from the document
                         //      but for now, we can leave this as a silent update since the block list would be for internal handling for now
                     }
+                    ConversationUpdateKind::ChangeSettings { settings } => {
+                        conversation.settings = settings;
```
dariusc93, line 1333: Shouldnt need to include the setting since its already apart of `conversation` in the update.

```suggestion
```

dariusc93 (commented):
Everything looks good. Left one small comment but besides that it will be ready to merge

Issue #415, opened 2024-01-07
raygun: Set a restriction to reactions


dariusc93:
In the current implementations, reactions are unbound in itself, allowing for any length of text for the reactions, however leaving it unbound internally can lead to increase in memory if a larger reaction is set. Adding a check to restrict the length would help, but the question would be to what specific length since some implementations may use unicode, while others might use shortcode/cldr and translate that on the frontend. 

For the time being, I believe, 128 characters will work, and what could be done is that we can have this adjustable in the configuration on a frontend-by-frontend basis, however restrict it so that it will only receive reaction events when those lengths is less than or equal to the configured amount, while anything more it would get rejected.

Issue #171, opened 2023-02-20
Keystore per conversation


dariusc93:
While direct messaging works fine as is, group messaging does not scale all as it was expected to be. This is primarily due to how a single message is encrypted per recipient in a conversation, which while for 3 to 5 recipients may be fine, large numbers of members (eg 10+ to 50+) may not scale well, which would increase cpu processing time, memory usage as well as an increase in network usage due to how large the payload would be. A proposal here would be to go back to the original idea from several months where a conversation would have its own keystore which would store the keys of recipients generated for that conversation. 

Thought Process (Simplified):

##### Summary
Each user would supply a unique private key to the conversation that would allow another user who is decrypt their messages. Keys supplied are automatically pushed out to each user upon being apart of the conversation or any changes in the recipient list, if there is an implementation to change keys around. 

#### Conversation Creation
- `A` creates a conversation for `B`, `C`, and `D`. `A` generates a private key and stores it into a keystore assigned with the conversation id
- An event is transmitted over to `B`, `C`, and `D`, with `A` generated private key, along with the list of recipients. 
- `B`, `C`, `D` will see the newly creation conversation with `A` private key. It stores it into a keystore encrypted with their own private key, assigned with the conversation id and will generate a key, which is stored into the same keystore for the conversation as well as transmitting their key to `A` as well as to each other.

#### Sending a Message
- Recipient `A` sends a message through RayGun::send_message
- Key is fetched from keystore for `A` and encrypts message with key. 
- Encrypted message is sent to all recipients in the conversation.

#### Receiving a Message
- Recipients receive message from `A`. 
- Recipients fetches `A` key from the keystore and decrypts it and validate the message. 
- When fetching the message stored, fetch the key of the sender and decrypt the message before passing it along. 

#### Adding Recipient
- `A` adds recipient `E`, transmitting the list of recipients as well as private key from keystore, notifying current recipients of `E`
  - (Optional) `E` may also receive keystore from `A` for all available recipients.
- `E` receives event, saving the recipients as well as `A` private key for the conversation, and sending its generated key to to `A` and all recipients.
- Recipients, upon being notified about new recipient, will send their private key for the conversation to `E`

#### Removing Recipient
- `A` removes recipient `B`, notifying current recipients of `B` removal. 
- `B` receives event, destroying the conversation as well as the keys stored apart of the conversation. 
- Recipients received notification upon being notified about removal of `B`, marking `B` key as `stale`
[...]

Issue #125, opened 2023-01-04
Shard message list and messages in a conversation


dariusc93:
Blocks stored is limited to 1MB by default in ipfs (with it not exceeding 2MB for bitswap). This creates a problem as messages build up over time for the list of cid linking to the encrypted messages. Once the block exceeds the default limit, it will likely error as it would not be able to obtain that block with a higher size, but even if there was a way to increase the block size externally (which at the moment that isnt an option), the issue there is that we wont be able to transmit the block over bitswap as it would exceeds the spec limit, which makes it hard to transfer the block to another node (eg for offline or backup). With this, what we should eventually do is shard the dag containing the list of cids as it begins to increase its size to have it point to an alternative dag containing additional messages. This would allow us to stay below the block limit as messages in a conversation begins to increase, as well as if the message itself (which may require splitting if it exceeds a specific size as the recipient list increases) but be able to keep things linked together when fetching or searching the messages later on. 

Scenarios:

* ~~A conversation with 50 recipients that has a single message containing 2048 character message would produce an approx 1.4MB block.~~

* A conversation with a single recipient (2 recipients in the conversation list) with 4250 messages, produce an approx 935KB block. This exclude encrypted messages since the list contains a small metadata document showing timestamp, message and conversation id, and the cid pointing to the encrypted message).  

Note: 
1. Though future plans would reduce the data itself,  sharding the block would reduce the single block size of the message if recipients will play a part in the encrypted message. 
2. A large portion of this has been resolved with #171

dariusc93:
The best way to properly handle this is to keep track of the size of the block and once it comes close to 1MB (or any defined size if we decide to increase or decrease it) to link it to the next block while while referencing the previous block in the new block.

dariusc93:
Resolved in #437

Issue #45, opened 2022-10-11
Offline Messaging


dariusc93:
Just like #44, messages will only send when the other user is online and connected, however if the other user is offline, the message would be stored locally into a queue-like system and would send out when the message when the other comes online, but if one user goes offline and the other comes online, they would not receive the message until the other comes online.

Note: Just like the issue above, this should not an enforced system. If unable to connect to the service or if its down for whatever reason it should continue operating as it currently does while checking for availability to ensure users are still able to communicate regardless of third party status.

PR #437, opened 2024-01-23
refactor: Use MessageDocument directly and use document variants

dariusc93:
Previously, we would send a `Message`, which would be encrypted based on the shared for direct messaging or #171 for group messaging due to originally sending this object directly prior to previous refactoring, however this begin to appear redundant as we had additional metadata on `MessageDocument` that mimics what was stored on `Message` (eg pinning indication, creation/modified dates, etc), while also storing the encrypted variant on this document. Furthermore, because we will be having additional metadata such as attachments, reactions, etc., on `MessageDocument`, it only made sense to use it directly instead of storing and `Message` directly.

**What this PR does** 📖
- Use `MessageDocument` directly when sending or referring to the internal message, encrypting the message lines itself rather than `Message`.
- Store `MessageDocument` in a reference list instead of a set with each list containing a max of 500 messages before descending into a new list for the next set. 
- Implement resolving `MessageDocument` into `Message`, allowing us to resolve via bitswap or some other content discovery and fetch protocol when walking its dag. 
- Send nonce of an newly edit message so the `MessageDocument` on the recipient end would contain the same signature when encrypting. 
- Add an signature, reactions and attachments as apart of `MessageDocument`, removing `Message::signature` (this field was used internally and been planned to have removed)
- Remove redundant and unused spam "filter" (will be replaced in the future with a proper filter and scoring system)
- Use file attachment document for attachments
- Increase file attachment limit from 8 to 32, and remove thumbnail limits

- If peer sends document but disconnects before it is resolved, the message will not resolve and would be skipped. This will change in the future either by matching both conversation documents on connection or some announcement of the message to the intended parties, or in regards to #45 implementation that will come soon (which the document would be pinned on shuttle, allowing it to be fetched by the intended recipient later).

**Additional comments** 🎤
- This change allows us to properly and cleanly pin the message on shuttle, and reduce overhead regarding the document itself due to `Message` being encrypted. This relates to #125.
- This reduces the size of the document itself, which in turn, would return the size of what is transmitted, and cut back on some of the complexity.
- Reactions may eventually be encrypted in the near future as well on the same document due to the metadata exposure. This will be left as TBD until sometime after #415
- While an nonce is to be used once for encrypting and apart of decrypting (for the alg used for this purpose), it is safe to share (as the nonce is shared apart of the encrypted message anyway and is emitted to recipients while being encrypted itself), and is used to make sure the encrypted message matches what the sender (which is the editor) message would be so it would validate. However, this should *only* be used once. If the current message uses the same nonce that is shared, the message will not be updated and would be rejected on the receiver end (in the future we may track previously used nonce to prevent them from being reused on both senders and receiver side).


dariusc93:
Blocked until #444 is merged

PR #443, opened 2024-01-31
chore: Remove a redundant channel

zeenix:
**What this PR does** 📖

Instead of sending commands through a channel and receiving the response through a oneshot channel created just for the response, just make the async calls directly.

While this compiles just fine, many conversation tests are failing after this change. Most likely, these changes are only exposing an issue, that was not being exposed before.

Keeping it draft until I can figure out the issue.


dariusc93:
As discussed previously, the purpose of `Conversations` is to allow a task to own the data to create a consistent state where communication is done via channels to create, update or fetch that data. Additionally, it would allow more flexibility by allowing us add streams into the mix while polling in that task. 

See #444 and #445 (both are draft at the time of this comment)

PR #444, opened 2024-02-01
refactor: Split MessageStore logic into a separate task

dariusc93:
**What this PR does** 📖
- Split messaging logic from `MessageStore` into a separate task
- Combine messages from the conversation topic, events and req/res topic into a single stream instead of polling them in separate tasks
- Temporarily store messages awaiting key exchange and process them once the exchange completed.

**Additional comments** 🎤
- `Conversations` was originally used to create a consistent state regarding conversations and messages, allowing us to save and fetch data in a FIFO, preserving the order of operations and allow the state of the data to be consistent without creating a race. However, it has become more apparent that a lot of the logic from `MessageStore` could benefit from being moved into a single task, which would give more flexibility when it comes to polling streams and handling various of task while preserving the state of the conversations and messages. This PR moves the majority of the logic into that single task, which is communicated from `MessageStore` via channels.


extensions/warp-ipfs/src/store/message.rs
```diff
@@ -1,2369 +1,2837 @@
...
+            state,
+            emoji,
+        };
 
-        Ok(Conversation::from(&conversation))
+        self.publish(conversation_id, None, event, true).await
     }
 
-    pub async fn create_group_conversation(
+    pub async fn attach(
```
dariusc93, line 2315: Just a note: This function internals will likely change in a near future PR to cleanup parts that been there that is no longer needed.

zeenix (approved):

PR #445, opened 2024-02-03
refactor: Move FileStore logic into a separate task

### LINKED ISSUE #435: constellation: uploading files depending on each other
When trying to upload multiple files via `warp::constellation::Constellation::put` (via e.g. a loop)
the returning `ConstellationProgressStream` only ends when all of the files are finished with uploading.

Expected behaviour would be instead that each stream will end when their respective progress is done.

```rs
for file in files {
     match warp_storage.put(&file, &some_path).await {
            Ok(upload_progress) => {
                // Handle each upload on another thread
                let mut warp_storage = warp_storage.clone();
                let res = tx.clone();
                tokio::spawn(async move {
                   // handle upload_progress
                   // upload_progress only ends if all files are done uploading
                   future_foo(upload_progress).await;
                });
            }
            Err(error) => log::error!("Error when upload file: {:?}", error),
        }
}
```

dariusc93:
**What this PR does** 📖
- Moves FileStore logic into a separate task
- Change internal logic for `put_buffer`, `get_buffer`, and `sync_ref` to send a future via oneshot.
- Remove hashing due to it taking longer for the stream to complete when dealing with larger files.

- Resolves #435 

**Additional comments** 🎤
- Hashing will be readded later where it will hash a file or data concurrent to the process of storing the data in the local node
  - This will be similar to how we generate thumbnails in a blocking thread and yield to the task when fetching it later.
  - Even with this change, hashing large data may still take awhile under other conditions like debug profile, etc.

### DIFF
--- a/extensions/warp-ipfs/src/store/files.rs
+++ b/extensions/warp-ipfs/src/store/files.rs
@@ -13,7 +13,10 @@ use rust_ipfs::{
     Ipfs, IpfsPath,
 };
 
-use tokio_util::io::ReaderStream;
+use tokio_util::{
+    io::ReaderStream,
+    sync::{CancellationToken, DropGuard},
+};
 use tracing::{Instrument, Span};
 use warp::{
     constellation::{
@@ -38,6 +41,7 @@ pub struct FileStore {
     path: Arc<RwLock<PathBuf>>,
     config: config::Config,
     tx: mpsc::Sender<FileTaskCommand>,
+    _guard: Arc<DropGuard>,
 }
 
 impl FileStore {
@@ -95,14 +99,18 @@ impl FileStore {
         let mut index = task.index.clone();
         let path = task.path.clone();
         let config = task.config.clone();
-        let _span = span.clone();
+        let span = span.clone();
 
         let signal = Some(task.signal_tx.clone());
         index.rebuild_paths(&signal);
 
+        let token = CancellationToken::new();
+        let _guard = Arc::new(token.clone().drop_guard());
+
         tokio::spawn(async move {
             tokio::select! {
-                _ = task.run().instrument(_span) => {}
+                _ = task.run().instrument(span) => {}
+                _ = token.cancelled() => {}
             }
         });
 
@@ -111,6 +119,7 @@ impl FileStore {
             config,
             path,
             tx,
+            _guard,
         })
     }
 }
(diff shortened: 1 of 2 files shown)


extensions/warp-ipfs/src/store/files.rs
```diff
@@ -32,26 +36,12 @@ use crate::{
...
-    constellation_tx: EventSubscription<ConstellationEventKind>,
-
-    signal: futures::channel::mpsc::UnboundedSender<()>,
-
     config: config::Config,
-
-    span: Span,
-
-    signal_guard: Arc<Notify>,
+    tx: mpsc::Sender<FileTaskCommand>,
```
zeenix, line 43: would be good to have a more descriptive name.

extensions/warp-ipfs/src/store/files.rs
```diff
@@ -84,73 +74,420 @@ impl FileStore {
...
+                response: tx,
+            })
+            .await;
+        rx.await.map_err(anyhow::Error::from)?
+    }
+
+    pub async fn put_buffer(
+        &mut self,
+        name: impl Into<String>,
+        buffer: impl Into<Vec<u8>>,
```
zeenix, line 207: Do we really want to be generic here? We know we we always want these specific types that allocate so we can just take those types directly. Also, if the caller accidentally passes a `String` as an `&str`, there will be unnecessary allocation/copying.

A better generic here might be `impl Into<Box<str>>` and `impl Into<Box<[u8]>>`. That would mean we don't need to allocate/copy for sending it over the channel and in some cases, [the conversion wouldn't require allocation/copying even](https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-From%3CVec%3CT,+A%3E%3E-for-Box%3C%5BT%5D,+A%3E).

zeenix (changes requested):
Mostly looks good.

extensions/warp-ipfs/src/store/files.rs
zeenix: > tx should be obvious that its a sending channel

That part is clear. What's not obvious is what is being sent over this channel and that makes the code less readable IMO. The names you removed where better, although I'm not sure if those still apply.

dariusc93: > That part is clear. What's not obvious is what is being sent over this channel and that makes the code less readable IMO. The names you removed where better, although I'm not sure if those still apply.

If one is reading over the code, I would assume that `tx:  mpsc::Sender<FileTaskCommand>` would be noticed, which would indicate what is being sent. Even then, reading further would indicate what is being sent via channel. I dont really believe it makes it less readable though. 

> The names you removed where better, although I'm not sure if those still apply.
Those arent applicable in this case :)

zeenix (approved):

PR #448, opened 2024-02-07
feat: Add group convo setting for "all members can change name"

zeenix:
**What this PR does** 📖

We already had the needed plumbing for group settings with "allow everyone to add participants" so we just add another boolean setting for "allow everyone to change conversation name".

### DIFF
--- a/extensions/warp-ipfs/examples/messenger.rs
+++ b/extensions/warp-ipfs/examples/messenger.rs
@@ -320,6 +320,7 @@ async fn main() -> anyhow::Result<()> {
                         CreateGroupConversation(name, did_keys, open) => {
                             let mut settings = GroupSettings::default();
                             settings.set_members_can_add_participants(open);
+                            settings.set_members_can_change_name(open);
                             if let Err(e) = chat.create_group_conversation(
                                 Some(name.to_string()),
                                 did_keys,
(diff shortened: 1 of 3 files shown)


extensions/warp-ipfs/examples/messenger.rs
```diff
@@ -369,6 +370,7 @@ async fn main() -> anyhow::Result<()> {
                             let topic = *topic.read();
                             let mut settings = GroupSettings::default();
                             settings.set_members_can_add_participants(open);
+                            settings.set_members_can_change_name(open);
```
dariusc93, line 373: This should probably be under its own command to make it more consistent with the rest of the code

dariusc93 (changes requested):
LGTM. Left a comment and can we update the test to add this in also?

extensions/warp-ipfs/examples/messenger.rs
zeenix: Well, we opted for the very generic "open" concept here so it would be strange to put it in a separate command. We can think about splitting this "open" concept into its individual parts in a separate PR for sure, though. We'll want to change the name of `--open` option of `create-group` as well then and add another one for this.

zeenix:
> LGTM. Left a comment and can we update the test to add this in also?

Right. I'll add a test.

dariusc93 (approved):

PR #453, opened 2024-02-20
chore: Cleanup & refactor Conversation types

zeenix:
**What this PR does** 📖

* chore: Turn `warp::raygun::Conversation` from struct to an enum. This makes sense and eliminates the disconnect between conversation type and conversation settings.
* creates a builder-pattern API for setting `*Conversation` types.
* Replaces `warp_ipfs::store::conversation::ConversationDocument` fields that are duplicated with `warp::raygun::Conversation`, with a `conversation` field of type `warp::raygun::Conversation`.
* Remove `warp::raygun::ConversationType`.

### DIFF
--- a/extensions/warp-ipfs/src/store/conversation.rs
+++ b/extensions/warp-ipfs/src/store/conversation.rs
@@ -14,9 +14,9 @@ use warp::{
     crypto::{cipher::Cipher, did_key::CoreSign, DIDKey, Ed25519KeyPair, KeyMaterial, DID},
     error::Error,
     raygun::{
-        Conversation, ConversationSettings, ConversationType, DirectConversationSettings,
-        GroupSettings, Message, MessageOptions, MessagePage, MessageReference, Messages,
-        MessagesType,
+        Conversation, ConversationSettings, ConversationType, DirectConversation,
+        DirectConversationSettings, GroupConversation, GroupSettings, Message, MessageOptions,
+        MessagePage, MessageReference, Messages, MessagesType,
     },
 };
 
@@ -34,18 +34,10 @@ pub enum ConversationVersion {
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq)]
 pub struct ConversationDocument {
-    pub id: Uuid,
+    #[serde(flatten)]
+    pub conversation: Conversation,
     #[serde(default)]
     pub version: ConversationVersion,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub creator: Option<DID>,
-    pub created: DateTime<Utc>,
-    pub modified: DateTime<Utc>,
-    pub conversation_type: ConversationType,
-    pub settings: ConversationSettings,
-    pub recipients: Vec<DID>,
     pub excluded: HashMap<DID, String>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub restrict: Vec<DID>,
@@ -59,27 +51,27 @@ pub struct ConversationDocument {
 
 impl Hash for ConversationDocument {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.id.hash(state)
+        self.conversation.hash(state)
     }
 }
 
 impl PartialEq for ConversationDocument {
     fn eq(&self, other: &Self) -> bool {
-        self.id == other.id
+        self.conversation.eq(&other.conversation)
     }
 }
 
 impl ConversationDocument {
     pub fn id(&self) -> Uuid {
-        self.id
+        self.conversation.id()
     }
 
     pub fn name(&self) -> Option<String> {
-        self.name.clone()
+        self.conversation.name().cloned()
     }
 
     pub fn topic(&self) -> String {
-        format!("{}/{}", self.conversation_type, self.id())
+        format!("{}/{}", self.conversation.conversation_type(), self.id())
     }
 
     pub fn event_topic(&self) -> String {
@@ -111,7 +103,8 @@ impl ConversationDocument {
             })
             .collect::<Vec<_>>();
 
-        self.recipients
+        self.conversation
+            .recipients()
             .iter()
             .filter(|recipient| !valid_keys.contains(recipient))
             .cloned()
@@ -120,46 +113,22 @@ impl ConversationDocument {
 }
 
 impl ConversationDocument {
-    #[allow(clippy::too_many_arguments)]
     pub fn new(
         did: &DID,
-        name: Option<String>,
-        mut recipients: Vec<DID>,
+        mut conversation: Conversation,
         restrict: Vec<DID>,
-        id: Option<Uuid>,
-        conversation_type: ConversationType,
-        settings: ConversationSettings,
-        created: Option<DateTime<Utc>>,
-        modified: Option<DateTime<Utc>>,
-        creator: Option<DID>,
         signature: Option<String>,
     ) -> Result<Self, Error> {
-        let id = id.unwrap_or_else(Uuid::new_v4);
-
-        if !recipients.contains(did) {
-            recipients.push(did.clone());
-        }
-
-        if recipients.is_empty() {
-            return Err(Error::CannotCreateConversation);
+        if !conversation.recipients().contains(did) {
+            conversation = conversation.add_recipient(did.clone());
         }
 
         let messages = None;
         let excluded = Default::default();
 
-        let created = created.unwrap_or(Utc::now());
-        let modified = modified.unwrap_or(created);
-
         let mut document = Self {
-            id,
             version: ConversationVersion::V1,
-            name,
-            recipients,
-            creator,
-            created,
-            modified,
-            conversation_type,
-            settings,
+            conversation,
             excluded,
             messages,
             signature,
@@ -171,7 +140,7 @@ impl ConversationDocument {
             document.verify()?;
         }
 
-        if let Some(creator) = document.creator.as_ref() {
+        if let Some(creator) = document.conversation.creator() {
             if creator.eq(did) {
                 document.sign(did)?;
             }
@@ -185,7 +154,7 @@ impl ConversationDocument {
         recipients: [DID; 2],
         settings: DirectConversationSettings,
     ) -> Result<Self, Error> {
-        let conversation_id = Some(super::generate_shared_topic(
+        let conversation_id = super::generate_shared_topic(
             did,
             recipients
                 .iter()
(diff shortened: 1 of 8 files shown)


warp/src/raygun/mod.rs
```diff
@@ -334,127 +334,262 @@ impl Ord for MessagePage {
...
+    }
+
+    pub fn settings(&self) -> ConversationSettings {
+        match self {
+            Conversation::Direct(direct) => ConversationSettings::Direct(direct.settings),
+            Conversation::Group(group) => ConversationSettings::Group(group.settings),
+        }
+    }
+
+    pub fn recipients(&self) -> &Vec<DID> {
```
dariusc93, line 375: nit
```suggestion
    pub fn recipients(&self) -> &[DID] {
```

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -34,18 +34,10 @@ pub enum ConversationVersion {
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq)]
 pub struct ConversationDocument {
-    pub id: Uuid,
+    #[serde(flatten)]
+    pub conversation: Conversation,
```
dariusc93, line 38: Im not sure about this part. Including `Conversation` in this document would allow reusing the fields from there instead of duplicating them here, but it also makes it hard to separate the internals. Eg, if we wish to use a `Cid` to represent the `recipients` list (unlikely, but could also be possible). then this design would make it impossible as `Conversation` should be agnostic and would be impossible to retrieve that data from `Conversation`. Another example would be if we were to include an image to `Conversation` (eg like a banner, which is may be a feature for groups, and eventually communities), we would have it use a `Cid` internally while externally, it would be in bytes on `Conversation`, so its hard to truly say if this should be here internally or not.

warp/src/raygun/mod.rs
```diff
@@ -334,127 +334,262 @@ impl Ord for MessagePage {
     }
 }
 
-#[derive(Debug, Hash, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Display)]
-#[serde(rename_all = "lowercase")]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Display)]
+#[serde(rename_all = "snake_case")]
```
dariusc93, line 338: Should this be untagged if its flatten internally?

warp/src/raygun/mod.rs
```diff
@@ -334,127 +334,262 @@ impl Ord for MessagePage {
...
     }
+}
 
-    pub fn set_name(&mut self, name: Option<String>) {
-        self.name = name;
+#[derive(Debug, Clone, Serialize, Deserialize, Eq)]
+pub struct DirectConversation {
+    #[serde(flatten)]
+    common: ConversationCommon,
+    settings: DirectConversationSettings,
```
dariusc93, line 472: As mentioned, I do feel this should be removed since direct conversations would not really have any permissions or settings of sort.

dariusc93 (changes requested):
I would have to think on this more because this would be a breaking change but also because of how the this is linking the external object to an internal, which I can see causing problems in the future. Furthermore, while I dont mind it for construction of `Conversation`, im not sure if the use-case completely fits to use `set_*` methods where it would return itself since I can also imagine it being confusing to somebody looking at it. Ill be looking over this more very soon just to piece together more thoughts for going forward.

extensions/warp-ipfs/src/store/conversation.rs
zeenix: > if we wish to use a `Cid` to represent the `recipients` list (unlikely, but could also be possible). then this design would make it impossible as `Conversation` should be agnostic and would be impossible to retrieve that data from `Conversation`.

How would that work with the current design? You still have recipients in the `Conversation` struct. Would recipients in Conversation be just ignored?

> Another example would be if we were to include an image to `Conversation` (eg like a banner, which is may be a feature for groups, and eventually communities), we would have it use a `Cid` internally while externally, it would be in bytes on `Conversation`, so its hard to truly say if this should be here internally or not.

Same question here.

---

Thing is that this change doesn't commit us to eliminate all duplication forever here. If the need arises to duplicate any fields (even the existing ones), then we can still do that.

zeenix: Also I think the representation of concepts should all go in the generic part (`warp`) and not the other way around. We can always make it super generic with traits and associated types, which would allow warp-ipfs to choose any type it wants as long as it satisfies certain trait bounds.

warp/src/raygun/mod.rs
zeenix: > As mentioned, I do feel this should be removed since direct conversations would not really have any permissions or settings of sort.

As long as we're certain of this but I'd like this to go in a separate PR. I'm doing a lot in here, as it is. :)

warp/src/raygun/mod.rs
zeenix: > Should this be untagged if its flatten internally?

That's up to the user if it's flattened or not. warps-ipfs is flattening it but we shouldn't assume that here.

zeenix:
> I would have to think on this more because this would be a breaking change

Breaking in the sense of crate API or IPFS? The former should be non-issue for us since Uplink links to specific commits. As for the latter, I tried to keep compatibility by flatting the struct when (de)serialization. Did I miss something?

zeenix:
> I dont mind it for construction of `Conversation`, im not sure if the use-case completely fits to use `set_*` methods

The builder pattern is quite common in the Rust world for setting fields that are optional to set but often need to be set at construction time. You can see it in action in a [few places in this PR](https://github.com/Satellite-im/Warp/pull/453/files#diff-a72e37f32768a8a57aca52ebcac90b6e49f612da34d6afacf2e1f8ab2cc0fe6dR186) in warp itself already. W/o the builder pattern, you'll either have to set the fields in separate statements (more writing) or pass many optional fields to the constructor (which can easily trigger the very legitimate clippy lint about too many args).

Besides, there are also `_mut` getters for cases where builder patter isn't great to use.

warp/src/raygun/mod.rs
zeenix: Oh, I see what you mean. hmm.. this will break the serialized format for sure. Let me think about it..

zeenix: I actually can't think of a good solution here since [untagged repr](https://serde.rs/enum-representations.html#untagged) means serde will have to do a trial-and-error when deserializing and it can get it wrong. We can use the internally-tagged rep though but that's still breaking the serialized form. 

Give that:
1. we have added fields in these structs and serialized formats recently.
2. this PR removes a field (`conversation_type`).

Do we really need to worry about the serialized format changing? 🤔

extensions/warp-ipfs/src/store/conversation.rs
dariusc93: > How would that work with the current design? You still have recipients in the Conversation struct. Would recipients in Conversation be just ignored?

When calling `RayGun::get_conversation` (or `RayGun::list_conversations`), internally we can resolve the recipients list if it was pointing to a `Cid` vs just converting from `ConversationDocument` to `Conversation` (as it does now). Under this setup, we have `Conversation` as a field in `ConversationDocument`  and if we were to have `recipients: Cid` (or some field representing pointing to a list of members in a conversation`, that would render the conversation field redundant.

Same if `Conversation` was to have an image supplied to it. On `ConversationDocument`, we may have a field that points to that image that is resolved when calling one of the functions above. While we could mark those fields as default and skip them when serializing, it still will appear to be redundant nonetheless.

warp/src/raygun/mod.rs
dariusc93: It can be done in a separate PR

warp/src/raygun/mod.rs
dariusc93: Could probably get by with using `#[serde(tag = "type")]` 

> Do we really need to worry about the serialized format changing?

Since this PR is using `conversations` in `ConversationDocument`, we would want to think about the formatting too since not all `serde` implementation will have the same representation and could easily error when serializing/deserializing or break (part of the reason why things are decoupled from what is shown in `warp` vs `warp-ipfs`, granted there are still some items that need to be resolved like `Message` being used internally, which #437 will also resolve).

zeenix: > Could probably get by with using `#[serde(tag = "type")]`

That's the internally-tagged repr and as I wrote, that will still break the serialized format. 

> could easily error when serializing/deserializing

How so? I don't see the connection between failures in (de)serialization here. `ConversationDocument` even shares the `Conversation` field types, e.g `ConversationSettings` and `ConversationType` currently. If there was a connection, it would equally apply to these duplicated fields.

> or break

As I wrote:

> 1. we have added fields in these structs and serialized formats recently.

i-e we do break the serialized format whenever we need to, it seems.

extensions/warp-ipfs/src/store/conversation.rs
zeenix: Thanks for explaining. Could you please also address my previous comment?

dariusc93: > Also I think the representation of concepts should all go in the generic part (warp) and not the other way around. We can always make it super generic with traits and associated types, which would allow warp-ipfs to choose any type it wants as long as it satisfies certain trait bounds.

Though would be outside of the scope of this PR (and probably will be done in a far later date for a full overhaul of the api), this would not really solve the overall issue since `ConversationDocument` use `Cid` that points to specific data (which at the time of this comment is used for messages, but would expand out to other items as well). It would not make sense to use `Conversation` in `ConversationDocument` in that case because we would still have fields internally and externally. In my opinion, it would be best to leave `ConversationDocument` as is and map out as needed.

dariusc93:
> The builder pattern is quite common in the Rust world for setting fields that are optional to set but often need to be set at construction time. You can see it in action in a [few places in this PR](https://github.com/Satellite-im/Warp/pull/453/files#diff-a72e37f32768a8a57aca52ebcac90b6e49f612da34d6afacf2e1f8ab2cc0fe6dR186) in warp itself already. 

Int his case, it sounds like it would be best to do this for `ConversationDocument` than `Conversation` (and other internal items), which I am all for (but can be done in a separate PR) :). 

> W/o the builder pattern, you'll either have to set the fields in separate statements (more writing) or pass many optional fields to the constructor (which can easily trigger the very legitimate clippy lint about too many args).

For `Conversation` itself, there would not really be a whole lot of arguments for a single function since it would be mostly setter/getter (which is used for constructing it for sending out to the API).

dariusc93:
Overall, I do feel we should hold off on this change for now. We could probably just make the changes externally so `Conversation` is an enum, and map things out accordingly from there, although we could also probably retain `ConversationType` to be use for some variant method of `Conversation::type`. The other option, which could be done in a separate PR, would be to remove the direct settings altogether and just retain the current settings specifically for groups.

zeenix:
Closing this. Will create a new PR that doesn't embed Conversation into ConversationDocument.

PR #458, opened 2024-02-25
refactor: Convert `Conversation` into a trait

### LINKED ISSUE #457: raygun: Convert `Conversation` into a trait
Currently, `Conversation` is a struct that contains data about the conversation. This may pertain about the name, list of members, etc. This was originally done since having `Conversation` as a trait previously raised some concern about the object safety as well as how to structure things safely across FFI, wasm (when wasm had some focused on at the time), however after some thought based on #453, I do believe we could try to implement `Conversation` as a trait instead as a struct. 

Design concept:

```rust
pub trait Conversation {
        /// ID of the converation in Uuid format
        fn id(&self) -> Uuid;

        /// Name of the conversation, if any
        fn name(&self) -> Option<String>;

        /// Time of when the conversation was created
        fn created(&self) -> DateTime<Utc>;

        /// Last time the conversation was modified or updated
        fn modified(&self) -> DateTime<Utc>;

        /// The conversation type
        /// Currently: Direct, Group
        fn conversation_type(&self) -> ConversationType;

        /// Conversation settings
        fn settings(&self) -> ConversationSettings;

        /// List of members to the conversation
        fn members(&self) -> Vec<DID>;
}
```

(Full concept)
```rust
pub trait ConversationUpdate {
        /// Update conversation name
        /// Note: This will only update the group conversation name
        async fn update_name(&mut self, name: &str) -> Result<(), Error>;

        /// Add a member to the conversation
        async fn add_member(&mut self, did_key: &DID) -> Result<(), Error>;

        /// Remove a member from the conversation
        async fn remove_member(&mut self, did_key: &DID) -> Result<(), Error>;

        /// Update conversation settings
        async fn update_settings(&mut self, settings: ConversationSettings) -> Result<(), Error>;
}

pub trait ConversationMessage {
        /// Retrieve a message from the conversation
        async fn get_message(&self, message_id: Uuid) -> Result<Message, Error>;

        /// Retrieve a message reference from the conversation
        async fn get_message_reference(&self, message_id: Uuid) -> Result<MessageReference, Error>;

        /// Retrieve messages from the conversation
        async fn get_messages(&self, options: MessageOptions) -> Result<Messages, Error>;

        /// Retrieve message references from the conversation
        async fn get_message_references(&self, options: MessageOptions) -> Result<BoxStream<'static, MessageReference>, Error>;

        /// Get a number of messages in the conversation
        async fn get_message_count(&self, message_id: Uuid) -> Result<usize, Error>;

        /// Get a status of a message in the conversation
        async fn message_status(&self, message_id: Uuid) -> Result<MessageStatus, Error>;

        /// Sends a message to a conversation.
        async fn send(&mut self, message: Vec<String>) -> Result<(), Error>;

        /// Edit an existing message in a conversation.
        async fn edit(
            &mut self,
            message_id: Uuid,
            message: Vec<String>,
        ) -> Result<(), Error>;

        /// Delete message from a conversation
        async fn delete(
            &mut self,
            message_id: Option<Uuid>,
        ) -> Result<(), Error>;

        /// React to a message
        async fn react(
            &mut self,
            message_id: Uuid,
            state: ReactionState,
            emoji: String,
        ) -> Result<(), Error>;

        /// Pin a message within a conversation
        async fn pin(
            &mut self,
            message_id: Uuid,
            state: PinState,
        ) -> Result<(), Error>;

        /// Reply to a message within a conversation
        async fn reply(
            &mut self,
            message_id: Uuid,
            message: Vec<String>,
        ) -> Result<(), Error>;

        /// Send an event to a conversation
        async fn send_event(&mut self, event: MessageEvent) -> Result<(), Error>;

        /// Cancel event that was sent, if any.
        async fn cancel_event(&mut self, event: MessageEvent) -> Result<(), Error>;
}

pub trait ConversationAttachment {
    /// Send files to a conversation.
    /// If no files is provided in the array, it will throw an error
    async fn attach(
        &mut self,
        message_id: Option<Uuid>,
        path: Vec<Location>,
        messages: Vec<String>,
    ) -> Result<AttachmentEventStream, Error>;

    /// Downloads a file that been attached to a message
    /// Note: Must use the filename associated when downloading
    async fn download(
        &self,
        message_id: Uuid,
        file: String,
        dest: PathBuf,
    ) -> Result<ConstellationProgressStream, Error>;

    /// Stream a file that been attached to a message
    /// Note: Must use the filename associated when downloading
    async fn download_stream(
        &self,
        message_id: Uuid,
        file: &str,
    ) -> Result<BoxStream<'static, Result<Vec<u8>, Error>>, Error>;
}

pub trait ConversationStream {
    /// Subscribe to an stream of events from the conversation
    async fn get_conversation_stream(&mut self) -> Result<MessageEventStream, Error>;
}
```

Having the trait designed this way allows us to migrate messaging logic to be specifically tied to the `Conversation` trait itself, removing the need to supply a `Uuid` thats representing the id to the conversation to the function itself. 

### Internals
Internally, we can have a struct that contains a clone of the task with metadata pretaining to the set of methods in `Conversation`, which would allow us to avoid making async calls. We will map each function to a specific internal function while also utilizing the `Uuid` specific to that conversation. 

When the conversation is deleted (either directly or indirectly), we will still allow the metadata stored for existing objects to be supplied to those methods, however they will no longer be updated. For the remaining methods, they will yield `Error::InvalidConversation` error (or some variant) showing that the conversation no longer exist.

We can deprecate the existing functions while defaulting raygun methods to the respected `Conversation` methods.  

### Pros
- (Internally, within warp-ipfs, may vary on other implementation) Information like name, list of members, etc., are updated internally, allowing one to make calls to those functions without needing to obtain a new conversation 
- Can pass around individual objects referencing the conversation itself, and be able to take actions and handle specific events related to that conversation

### Cons
- When conversation is closed or deleted, `Conversation` object may still be left hanging around, preserving the metadata. While this is no problem since internally it is disconnected from the conversation and would no longer be updated, it may cause some to believe the conversation is still active. 
  - We could return `Result<T, warp::error::Error>` to the data getter methods, however this would break compatibility with the existing `Conversation` getter api to such methods.
  - We could panic when those getter methods are called to a closed conversation since realistically, upon the conversation being deleted and the event being emitted, the frontend should handle the cleanup of those references, however this would be a race as one thread may access the api while the conversation is being deleted but prior to the api being emitted.
- Requires maintaining `Conversation` for messaging logic unless one calls `RayGun::get_conversation` and executing the method right away and dropping the trait object.

Note:
- `Conversation` should be object safe (which the implementation above is)
- We can only implement `Conversation` trait instead and leave the remaining logic intact, allowing for more compatibility but may lose out on the flexability of executing messaging logic directly on the trait object. This would reduce the refactoring aspect besides renaming the method from `Conversation::recipients` (old) to `Conversation::members` (new).

dariusc93:
**What this PR does** 📖
- Refactors `Conversation` into a trait object
- Deprecate current raygun messaging logic, while defaulting to `Conversation` messaging logic for compatibility.

Resolves #457 

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-ipfs/src/lib.rs
+++ b/extensions/warp-ipfs/src/lib.rs
@@ -46,10 +46,9 @@ use warp::crypto::keypair::PhraseType;
 use warp::crypto::zeroize::Zeroizing;
 use warp::module::Module;
 use warp::raygun::{
-    AttachmentEventStream, Conversation, ConversationSettings, EmbedState, GroupSettings, Location,
-    Message, MessageEvent, MessageEventStream, MessageOptions, MessageReference, MessageStatus,
-    Messages, PinState, RayGun, RayGunAttachment, RayGunEventKind, RayGunEventStream, RayGunEvents,
-    RayGunGroupConversation, RayGunStream, ReactionState,
+    AttachmentEventStream, Conversation, GroupSettings, Location, MessageEvent, MessageEventStream,
+    RayGun, RayGunAttachment, RayGunEventKind, RayGunEventStream, RayGunEvents,
+    RayGunGroupConversation, RayGunStream,
 };
 use warp::tesseract::{Tesseract, TesseractEvent};
 use warp::{Extension, SingleHandle};
@@ -1355,8 +1354,10 @@ impl IdentityInformation for WarpIpfs {
 
 #[async_trait::async_trait]
 impl RayGun for WarpIpfs {
-    async fn create_conversation(&mut self, did_key: &DID) -> Result<Conversation, Error> {
-        self.messaging_store()?.create_conversation(did_key).await
+    async fn create_conversation(&mut self, did_key: &DID) -> Result<Box<dyn Conversation>, Error> {
+        let store = self.messaging_store()?;
+        let id = store.create_conversation(did_key).await?;
+        store.get_conversation(id).await
     }
 
     async fn create_group_conversation(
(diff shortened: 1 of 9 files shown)


extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -624,26 +623,26 @@ impl ConversationDocument {
...
+//         conversation.set_name(document.name.clone());
+//         conversation.set_creator(document.creator.clone());
+//         conversation.set_conversation_type(document.conversation_type);
+//         conversation.set_recipients(document.recipients());
+//         conversation.set_created(document.created);
+//         conversation.set_settings(document.settings);
+//         conversation.set_modified(document.modified);
+//         conversation
+//     }
+// }
```
zeenix, line 645: This doesn't make any sense after `Conversation` becomes a trait so there is no reason to keep it commented out. Just delete it.

extensions/warp-ipfs/src/lib.rs
```diff
@@ -1355,158 +1354,36 @@ impl IdentityInformation for WarpIpfs {
...
-
-    async fn update_conversation_settings(
-        &mut self,
-        conversation_id: Uuid,
-        settings: ConversationSettings,
-    ) -> Result<(), Error> {
-        self.messaging_store()?
-            .update_conversation_settings(conversation_id, settings)
-            .await
+    async fn list_conversations(&self) -> Result<Vec<Box<dyn Conversation>>, Error> {
```
zeenix, line 1385: can't we just return a `&[Box<..>]` here or `impl Iterator<Item = Box<dyn Conversation>> + '_` here? 🤔

extensions/warp-ipfs/src/store/message.rs
```diff
@@ -1149,19 +1189,20 @@ impl ConversationTask {
...
-            .find(|conversation| {
-                conversation.conversation_type == ConversationType::Direct
-                    && conversation.recipients().contains(did)
-                    && conversation.recipients().contains(&self.keypair)
-            })
-            .map(Conversation::from)
-        {
-            return Err(Error::ConversationExist { conversation });
-        }
+        //TODO
```
zeenix, line 1192: TODO for this PR, I hope?

extensions/warp-ipfs/tests/direct.rs
```diff
@@ -313,11 +314,11 @@ mod test {
         let message_b = tokio::time::timeout(Duration::from_secs(60), async {
             loop {
                 if let Some(MessageEventKind::MessageReceived {
-                    conversation_id,
+                    conversation_id: _,
```
zeenix, line 317: I think you can just skip everything with `..` here.
```suggestion
                    ..
```

warp/src/raygun/mod.rs
```diff
@@ -1118,11 +1195,13 @@ pub trait RayGunStream: Sync + Send {
 
 #[async_trait::async_trait]
 pub trait RayGunEvents: Sync + Send {
+    #[deprecated(note = "Use `Conversation::send_event`")]
```
zeenix, line 1198: I'm not sure why bother with deprecation. Do we have any external users and have we promised any API stability guarantees? I don't think we should unless we publish the crates on crates.io.

extensions/warp-ipfs/src/store/conversation.rs
dariusc93: Left there unintentionally

extensions/warp-ipfs/src/store/message.rs
dariusc93: It is

warp/src/raygun/mod.rs
dariusc93: It gives a chance to frontend implementation (specifically, uplink) to determine how they wish to proceed in the change while indicating to them that the functions are deprecated

extensions/warp-ipfs/tests/direct.rs
dariusc93: left intentionally (since i didnt complete the whole test but meant to compare conversation id but do keep in mind this pr is still in a draft state and just putting the concept out there)

extensions/warp-ipfs/src/lib.rs
dariusc93: If I recall doing that would cause the trait not to object safe

zeenix: I think there is a different issue with doing that: [Send and Sync](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=90dd7c07c6d58423f71cff1afd84ff44).

extensions/warp-ipfs/tests/direct.rs
zeenix: sure. I noticed half way through both PRs that they're still drafts. :)

warp/src/raygun/mod.rs
zeenix: I really don't think that's needed here. The frontend is also us so if there is a doubt on the new API to be problematic for Uplink, we just make the changes there and see if it works as expected. If not, we rethink/redo the warp changes. Deprecation is meant for public APIs and it would make sense even across teams in the same organisation but that's not the case here. It's just us on both ends. :)

extensions/warp-ipfs/src/lib.rs
dariusc93: > Do we need object safety WarpIpfs trait?
It is preferred at this point to make sure there is object safety.

warp/src/raygun/mod.rs
dariusc93: While its not 100% needed to be deprecated, this type of change would be a major breaking change for uplink if they are removed right away. Keeping the api intact temporarily while they sort out how they wish to handle it internally since with this design it is not required to make repeated calls to `RayGun::get_conversation` but can make a single call and handle things from there based on any events emitted, etc. If this PR is decided on, the PR I would make on uplink would only do essential update to the code to ensure functionality while making note of the changes needed for somebody else to take care of that would fix their design on that platform. 

> The frontend is also us so if there is a doubt on the new API to be problematic for Uplink, we just make the changes there and see if it works as expected. If not, we rethink/redo the warp changes

Keep in mind, ideally warp itself may be available for anybody to implement, not just uplink (which there have been a couple of cases ive seen, though not sure of those implementation status). While we want to make sure the design make sense, we shouldnt have to go back and forth changing stuff warp just so it would fit an idea on one implementation (ie uplink) that may not be sound in other implementation. Always best to find the middle ground whenever possible. Just an FYI :)

zeenix: Thanks for explaining your arguments for this in detail. I appreciate that but I'm still not convinced this is worth it. Ideally we should never use deprecated API (not only in Uplink but anywhere) and not having to add "allow(deprecated)" that are easy to forget and ending up later using other deprecated API unintentionally.

> Keep in mind, ideally warp itself may be available for anybody to implement

That's not relevant to us. We don't expose a public API nor we make a claim about its stability.

More importantly. We are breaking the API big time in this PR itself. If we want to provide any stability, we should do that then. This half-way care of API break, is not going to help anyone.

zeenix: To sumarize, I really think we either guarantee API stability and hence always deprecate existing API and follow the semver rules, OR we don't. I don't see any reason to do both.

PR #459, opened 2024-02-25
refactor: Convert `Conversation` into a trait (partial)

dariusc93:
**What this PR does** 📖
- Convert `Conversation` into a trait object

- Relates to #457 

- This PR, much like #458, is conceptual and still TBD for #457

**Additional comments** 🎤
- Unlike #458, this PR only supports having `Conversation` as the trait object while retaining the rest of the messaging logic, besides the breaking change between `Conversation::recipients` and `Conversation::members`
- This object attempts to remain as a getter functionality while attempting to avoid async calls, however similar to #458, we may eventually need to cache the resolved data of the `ConversationDocument` and map that out to this trait object if we wish to maintain the same flow without introducing async functions into the mix.
- Unlike #458, this PR only implements `Conversation`. It does not store its internal data inside of a lock, which means when when the conversation has an updated (eg adding/removing members, update to conversation name, etc), it would need to grab a new object by calling `RayGun::get_conversation` with the updated information

### DIFF
--- a/warp/src/raygun/mod.rs
+++ b/warp/src/raygun/mod.rs
@@ -344,118 +344,28 @@ pub enum ConversationType {
     Group,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Eq)]
-pub struct Conversation {
-    id: Uuid,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    name: Option<String>,
-    creator: Option<DID>,
-    created: DateTime<Utc>,
-    modified: DateTime<Utc>,
-    conversation_type: ConversationType,
-    settings: ConversationSettings,
-    recipients: Vec<DID>,
-}
+pub trait Conversation: Debug + DynClone + Send + Sync {
+    /// ID of the converation in Uuid format
+    fn id(&self) -> Uuid;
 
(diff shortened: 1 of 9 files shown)


warp/src/raygun/mod.rs
```diff
@@ -936,17 +846,17 @@ pub trait RayGun:
...
 
     /// Get an active conversation
-    async fn get_conversation(&self, _: Uuid) -> Result<Conversation, Error> {
+    async fn get_conversation(&self, _: Uuid) -> Result<Box<dyn Conversation>, Error> {
         Err(Error::Unimplemented)
     }
 
     /// List all active conversations
-    async fn list_conversations(&self) -> Result<Vec<Conversation>, Error> {
+    async fn list_conversations(&self) -> Result<Vec<Box<dyn Conversation>>, Error> {
```
zeenix, line 859: same question here as the other PR: can't we just return a `&[Box<..>]` here or `impl Iterator<Item = Box<dyn Conversation>> + '_` here? 🤔

dariusc93: Doing that would cause the trait not to be object safe if i recall (same comment on the other PR)

PR #462, opened 2024-02-29
feat: make messages send return their uuid

### LINKED ISSUE #461: return message id when sending
Currently for uplink when we send a message we also add a pending message to track to it. 
it would be nice if warps `warp_ipfs::store::message::MessageStore#send_message` would return the generated id (`Result<Uuid, Error>`) for that message instead of just returning a `Result<(), Error>` to make tracking it easier.
atm a pending message gets identified by its text + attachments instead of a uuid which makes it not really reliable for e.g. same text msg etc. (it does gets assigned a uuid too but there are some caveat to it irrc)

Flemmli97:
**What this PR does** 📖

Makes sending messages return their uuid for easier tracking

- Close #461

### DIFF
--- a/extensions/warp-ipfs/src/lib.rs
+++ b/extensions/warp-ipfs/src/lib.rs
@@ -1432,7 +1432,7 @@ impl RayGun for WarpIpfs {
             .await
     }
 
-    async fn send(&mut self, conversation_id: Uuid, value: Vec<String>) -> Result<(), Error> {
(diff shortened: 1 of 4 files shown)


dariusc93 (changes requested):
Thanks for the PR. Can you deal with the lint errors? Besides that everything LGTM

dariusc93 (approved):

zeenix (approved):

PR #463, opened 2024-03-01
feat: return uuid together with attachment stream

Flemmli97:
**What this PR does** 📖

Returns a uuid for a message with attachment together with the upload stream


dariusc93 (commented):
on line 2559 in `store/message.rs`, we should have this return back to returning `()` dont you think?

dariusc93 (approved):

PR #466, opened 2024-03-06
chore: Remove disconnect b/w Conversation* type-specific fields

zeenix:
**What this PR does** 📖

This PR remove the ConversationType fields in 2 structs and add getters that infer the type from the settings field. This way, it eliminates the disconnect between them and ensures that the type is never inconsistent with the settings.


dariusc93 (approved):
Not to happy about the removal of the conversation_type field since it does break the conversation representation when serialized. Furthermore, im not to sure on the use of `ConversationSettings` as the representation for `ConversationType` as `ConversationSettings` could possibly change in the future (eg to be more permission based for group conversations, etc)

PR #474, opened 2024-03-14
chore: Use a single lock in constellation index

dariusc93:
**What this PR does** 📖
- Reduce allocations in `File` and `Directory` by moving fields into a separate struct and wrap that struct using a single `Arc<RwLock<T>>` in its respected struct. 

**Additional comments** 🎤


dariusc93:
This will be redone in a future PR

PR #476, opened 2024-03-18
chore: Cleanup dependencies

dariusc93:
**What this PR does** 📖
- Use specific features from crates

**Additional comments** 🎤

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,16 +15,22 @@ resolver = "2"
 [workspace.dependencies]
 
 # Async and futures crates
-futures = { version = "0.3" }
+futures = { version = "0.3", default-features = false }
 async-trait = { version = "0.1" }
 async-stream = "0.3"
 async-broadcast = "0.5"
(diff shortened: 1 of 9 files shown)


zeenix:
In the end I couldn't find many candidates for feature disabling and also it turned out to be quite a boring task. 😆 I think this is good enough for now.

Cargo.toml
```diff
@@ -15,16 +15,22 @@ resolver = "2"
 [workspace.dependencies]
 
 # Async and futures crates
-futures = { version = "0.3" }
+futures = { version = "0.3", default-features = false }
```
dariusc93, line 18: Wouldnt this disable features that enables channels, use of sink/stream, and other utils from its underlining crates behind its default feature? If so then we should probably enable those features thats being used here.

zeenix: You're right. I think something else enables those features so the build is not broken. That's not something we should rely on though. It seems `std` feature enables all the stuff we need.

PR #477, opened 2024-03-18
chore: Drop unneeded dependencies

zeenix:
**What this PR does** 📖

Drop unneeded dependencies. In some cases the only use of a crate is replaced by the same/similar API from another crate we already depend on.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ resolver = "2"
 
 # Async and futures crates
 futures = { version = "0.3" }
-futures-timer = "3.0"
 async-trait = { version = "0.1" }
 async-stream = "0.3"
 async-broadcast = "0.5"
@@ -25,13 +24,8 @@ tokio-util = { version = "0.7" }
 tokio-stream = { version = "0.1" }
 
 # Crypto crates
-x25519-dalek = { version = "1" }
 ed25519-dalek = { version = "1" }
-curve25519-dalek = "3.2"
-blake2 = { version = "0.10" }
-sha1 = { version = "0.10" }
 sha2 = { version = "0.10" }
-sha3 = { version = "0.10" }
 hmac = { version = "0.12.0", default-features = false }
 digest = { version = "0.10" }
 aes-gcm = { version = "0.10" }
@@ -47,7 +41,6 @@ thiserror = "1.0"
 
 # Sync crates
 parking_lot = { version = "0.12" }
-cfg-if = "1.0"
 once_cell = "1.16"
 
 # Time crate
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -72,16 +65,15 @@ dyn-clone = "1.0"
...
-getrandom = { version = "0.2" }
 either = "1"
 void = "1"
 
-# ipfs dependency
-rust-ipfs = "0.11.0"
+#ipfs dependency
+rust-ipfs = { version = "0.11.0", default-features = false, features = [
+    "libp2p_bitswap",
+] }
```
dariusc93, line 76:
```suggestion
rust-ipfs = "0.11.0"
```

warp/src/crypto/mod.rs
```diff
@@ -160,6 +157,8 @@ impl From<ed25519_dalek::SecretKey> for DID {
 
 pub fn generate<const N: usize>() -> [u8; N] {
     let mut buf = [0u8; N];
-    getrandom::getrandom(&mut buf).unwrap();
+    for x in buf.iter_mut() {
+        *x = rand::random()
+    }
```
dariusc93, line 162: You could use `RngCore::fill_bytes` instead (would require importing `OsRng` and `RngCore`)

```suggestion
    OsRng.fill_bytes(&mut buf);
```

dariusc93 (changes requested):

Cargo.toml
zeenix: why? we only need the specific feature, no?

warp/src/crypto/mod.rs
zeenix: `OsRng` uses `getrandom` so it'll beat the point of this change.

Cargo.toml
dariusc93: The feature isnt needed here anymore (the default-features was not needed either).

zeenix: ok but why enable all features then?

warp/src/crypto/mod.rs
dariusc93: rand uses getrandom by default anyway the last I check (though getrandom may be more reliable to use for seeding), otherwise you need to disable the default features and use `std` and `std_rng` features and maybe change it to 

```rust
    let mut rng = rand::thread_rng();
    rng.fill_bytes(&mut buf);
```

Cargo.toml
dariusc93: It doesnt contain any default features.

zeenix: ah ok. :) That's the detail I missed.

warp/src/crypto/mod.rs
zeenix: > rand uses getrandom by default anyway the last I check

I did look at the code before my previous comment, and it doesn't. See [OsRng docs](https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html).

dariusc93: Looking at rand itself, `std` feature uses `getrandom`

See https://github.com/rust-random/rand/blob/0.8.5/Cargo.toml#L36

zeenix: Ah, you're right. Thanks.

dariusc93 (approved):

PR #480, opened 2024-03-20
feat: Return location for message attachments

Flemmli97:
**What this PR does** 📖

Makes it so the original location is also returned when sending a message with attachments to it. 
The returned stream is now a pair of (optional) Location + Progress

### DIFF
--- a/extensions/warp-ipfs/src/store/message.rs
+++ b/extensions/warp-ipfs/src/store/message.rs
@@ -2408,7 +2408,7 @@ impl ConversationTask {
             }
 
             for await (location, (progress, file)) in streams {
-                yield (Some(location), AttachmentKind::AttachedProgress(progress));
+                yield AttachmentKind::AttachedProgress(location, progress);
                 if let Some(file) = file {
                     attachments.push(file);
                 }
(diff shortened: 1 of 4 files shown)


extensions/warp-ipfs/src/store/message.rs
```diff
@@ -2294,7 +2294,7 @@ impl ConversationTask {
             let mut streams: SelectAll<_> = SelectAll::new();
 
             for file in files {
-                match file {
+                match file.clone() {
```
dariusc93, line 2297: You can borrow this instead

```suggestion
                match &file {
```

dariusc93 (changes requested):
If we are to use `Location`, maybe we should push that into `AttachmentKind::AttachedProgress` instead? That way we dont need to change the stream output itself but only `AttachmentKind`

Flemmli97:
> If we are to use `Location`, maybe we should push that into `AttachmentKind::AttachedProgress` instead? That way we dont need to change the stream output itself but only `AttachmentKind`



> If we are to use `Location`, maybe we should push that into `AttachmentKind::AttachedProgress` instead? That way we dont need to change the stream output itself but only `AttachmentKind`

idk why i thought that was used somewhere else which is why i initially didn't do it.
changed it now

dariusc93 (approved):

Issue #494, opened 2024-04-12
feat: Add wasm support


dariusc93:
Previously, there were wasm support at the interface level, which also exported out to javascript as well, though this increased some complexity. However, the wasm research and implementation was postponed due to a different focus. Currently, if we are to implement wasm, we should make sure that dependencies do support wasm, split/gate off any file IO and native logic (eg reading/writing files, blocking tasks, etc) and enable the appropriate features where needed for wasm support. 

Note:
- ~~This is blocked until `rust-ipfs` has wasm support for `warp-ipfs`~~
  - There is basic wasm support in `rust-ipfs` to attempt a PoC, however more changes will be needed upstream and there may be bugs.
- Not all browsers support wasm or may have different requirements towards specific API so this may need to be taken into account.
- This does not require any exporting to javascript to prevent/reduce complexity but instead focuses on the use of rust and allow those who uses the API to make their own wrappers or exportation to javascript according to their use-case.

dariusc93:
A couple of things to note:
- in-memory store is only supported at this time. 
- When targeting wasm, we lose out on the ability to listen on a specific address since we are not able to listen to addresses and ports from the browser (though in the far future there may be browser-to-browser with webrtc)
- We lose transports for tcp, quic, and would rely on webrtc and websocket transports, in which case may require using a relay to connect to other peers.
  - It is advised to use native implementation for mobile and desktop while for web will fall back to those transports
- Performance may vary and may not be up to par compared to native implementation

dariusc93:
Notes for Constellation:
- We should exclude `Constellation::get` and `Constellation::put` and instead rely on supplying a stream (`Constellation::put_stream` or `Constellation::get_stream`) or a buffer (`Constellation::put_buffer` or `Constellation::get_buffer`) when coming to wasm due to being unable to pull from the filesystem directly due to the limitation (and security aspect) of the browser.

dariusc93:
Notes for MultiPass:
- A buffer or stream should be used when uploading/updating the profile picture and banner since we are unable to pull directly from the filesystem.
- The crate `image` should be tested to determine if it supports being compiled to wasm via `wasm32-unknown-unknown` target

dariusc93:
Notes for RayGun:
- `RayGun::download_stream` should be used when downloading to a filesystem so the wasm implementation can handle downloading it to disk
[...]

PR #496, opened 2024-04-15
chore: fix some wasm compilation issues

ashneverdawn:
**What this PR does** 📖

fixes some of the compilation issues on wasm. 
Package `warp` now compiles, and package `warp-ipfs` almost compiles (1 error left) 
`tools/shuttle` now uses `LocalStorage` instead of `Tokio:fs` when compiled for wasm

#494 (partially)

**Additional comments** 🎤

### DIFF
--- a/tools/shuttle/src/store/root.rs
+++ b/tools/shuttle/src/store/root.rs
@@ -32,12 +32,31 @@ pub struct RootStorage {
 
 impl RootStorage {
     pub async fn new(ipfs: &Ipfs, path: Option<PathBuf>) -> Self {
+        
+        #[cfg(not(target_arch = "wasm32"))]
         let root_cid = match path.as_ref() {
             Some(path) => tokio::fs::read(path.join(".root_v0"))
                 .await
                 .map(|bytes| String::from_utf8_lossy(&bytes).to_string())
                 .ok()
                 .and_then(|cid_str| cid_str.parse().ok()),
+            None => None,
+        };
+
+        #[cfg(target_arch = "wasm32")]
+        let root_cid = match path.as_ref() {
+            Some(path) => path.to_str().and_then(|key| {
+                use gloo::storage::{LocalStorage, Storage};
+                let local_storage = LocalStorage::raw();
+                local_storage.get(key).ok().and_then(|key| {
+                    key.and_then(|key|{
+                        LocalStorage::get(key).ok().and_then(|bytes: Vec<u8> |{
+                            let cid_str = String::from_utf8_lossy(&bytes).to_string();
+                            cid_str.parse().ok()
+                        })
+                    })
+                })
+            }),
             None => None,
         };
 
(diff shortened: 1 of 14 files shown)


tools/shuttle/src/store/root.rs
```diff
@@ -41,6 +43,23 @@ impl RootStorage {
...
+                    key.and_then(|key|{
+                        LocalStorage::get(key).ok().and_then(|bytes: Vec<u8> |{
+                            let cid_str = String::from_utf8_lossy(&bytes).to_string();
+                            cid_str.parse().ok()
+                        })
+                    })
+                })
+            }),
+            None => None,
+        };
```
dariusc93, line 61: This wouldnt be needed since the root store is apart of the server-side code so we could use `#[cfg(not(target_arch = "wasm32"))]` on https://github.com/Satellite-im/Warp/blob/b574da4ab10cfd7bf38ec0e2b499232a2e8f8eff/tools/shuttle/src/lib.rs#L16-L18 which should allow shuttle to compile.

tools/shuttle/src/store/root.rs
```diff
@@ -139,10 +158,19 @@ impl RootInner {
...
+            #[cfg(not(target_arch = "wasm32"))] {
+                if let Err(e) = tokio::fs::write(path.join(".root_v0"), cid).await {
+                    tracing::error!("Error writing cid to file: {e}");
+                }
+            }
+            #[cfg(target_arch = "wasm32")] {
+                use gloo::storage::{LocalStorage, Storage};
+                if let Err(e) = LocalStorage::set(path.join(".root_v0").to_str().unwrap(), cid) {
+                    tracing::error!("Error writing cid to file: {e}");
+                }
```
dariusc93, line 170: ditto

warp/src/crypto/cipher.rs
```diff
@@ -1,6 +1,5 @@
 #![allow(clippy::result_large_err)]
 #[allow(unused)]
-#[cfg(not(target_arch = "wasm32"))]
```
dariusc93, line 3: Does `wasm32-unknown-unknown` support `io::Read` and `io::Write` now? We could probably move `std::io::ErrorKind` out into its own import (since I do know this would error if this attr is used for `ErrorKind`) and keep this so we are not including `Read` or `Write` it by mistake unless it is supported by wasm32 arch.

dariusc93 (changes requested):
Thanks for the PR. Left a comment but besides that LGTM

tools/shuttle/src/store/root.rs
ashneverdawn: done

tools/shuttle/src/store/root.rs
ashneverdawn: done

warp/src/crypto/cipher.rs
ashneverdawn: that said I separated ErrorKind as you suggested

dariusc93 (approved):

PR #499, opened 2024-04-18
chore: Add wasm-bindgen-futures, add example

dariusc93:
**What this PR does** 📖
- Adds wasm-bindgen-futures
- Add a `rt` module in `warp-ipfs` to add a alias between spawning tokio and wasm-bindgen task
- Use `web-time` as a drop-in replacement for `std::time::Instant`
- Disable rayon in image due to incompatibility with wasm target

Relates to #494 

**Additional comments** 🎤
- The `rt` module, while it supports spawning a task for tokio or wasm, does not return a handler for wasm side, so it is best to use for task that dont expect to return anything
- Example is added, which is based on `ipfs-identity`, which creates and update the identity. This does not use any networking bits at this time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,9 @@ hex = "0.4"
 libipld = { version = "0.16", features = ["serde-codec"] }
 bytes = "1"
 bincode = "1"
-image = "0.24"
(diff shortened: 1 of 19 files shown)


ashneverdawn (approved):
awesome! lgtm

PR #505, opened 2024-04-25
Chore: export tesseract to wasm

ashneverdawn:
**What this PR does** 📖

Exports tesseract to wasm, making it usable from the browser. Also includes example usage in js.

#494 (partially)

**Additional comments** 🎤


warp/src/tesseract/mod.rs
```diff
@@ -19,12 +20,16 @@ use parking_lot::RwLock;
 
 type Result<T> = std::result::Result<T, Error>;
 
+#[cfg(target_arch = "wasm32")]
+pub mod wasm;
```
dariusc93, line 24: Here instead of having a separate module, we could use `cfg_attr` to check if the target is wasm and if so, add `wasm_bindgen` respectfully and if needed we can split the impl into multiple impl with one for wasm and one for non-wasm target.

warp/src/tesseract/mod.rs
```diff
@@ -1,6 +1,7 @@
 #![allow(clippy::result_large_err)]
 use futures::{stream::BoxStream, StreamExt};
 use std::{collections::HashMap, fmt::Debug, sync::atomic::Ordering};
+use wasm_bindgen::prelude::*;
```
dariusc93, line 4:
```suggestion
#[cfg(target_arch = "wasm32")]
use wasm_bindgen::prelude::*;
```

www/tesseract.html
```diff
@@ -0,0 +1,72 @@
...
+  </script>
+  <body>
+    <h2>How to use this example:</h2>
+    <h3>1 - Make the sure wasm is compiled:</h3>
+    <pre>
+      wasm-pack build tools/wasm-example --target web --out-dir www/wasm/wasm-example
+    </pre>
+    <h3>2 - Check ouput in the console</h3>
+  </body>
+</html>
```
dariusc93, line 72: Can be done in a separate PR but the `www` directory should probably be moved to the wasm example directory in `tools`

warp/src/tesseract/wasm.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    pub fn load_from_storage(&self) -> Result<(), JsError> {
+        self.inner.load_from_storage().map_err(|e| e.into())
+    }
+}
+
+#[wasm_bindgen]
+pub struct Subscription {
+    inner: BoxStream<'static, TesseractEvent>,
+}
```
dariusc93, line 98: I take it that wasm-bindgen doesnt have anything similar to streams we could wrap around or if `wasm-streams` isnt suitable here?

warp/src/tesseract/wasm.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[wasm_bindgen]
+impl Subscription {
+    pub async fn next(&mut self) -> Option<TesseractEvent> {
+        self.inner.next().await
```
dariusc93, line 103: I think the problem here is that we would be continuously polling a stream and if that stream returns `None` (or `Poll::Ready(None)`), we should stop polling otherwise it would in many cases be considered as a UB (eg if tesseract completely drops (as in no other references), the sender used would also drop, causing the receiver to close too). If we are to go this route, we might want to see if there is a way for JS to break that loop so it no longer polls if it returns `None` from rust. Cannot recall the equiv translation that wasm-bindgen export would unless it just returns as a `null`, in which case we could probably try that. Ideally though, it would be best to try to use an asynciterator or maybe tap into the event listener from js so the events are subscribe there.

dariusc93 (changes requested):

warp/src/tesseract/mod.rs
ashneverdawn: done

warp/src/tesseract/mod.rs
ashneverdawn: done

www/tesseract.html
ashneverdawn: done

warp/src/tesseract/wasm.rs
ashneverdawn: no, not built in at least. Only has basic types like struct, bool, int, str, etc. `wasm-streams` didn't work as a return type exported to js, unfortunately.

ashneverdawn: In this case, the exact reason we need to wrap this is because of the `'static` lifetime. js doesn't understand lifetimes. But by wrapping this, it can live within the wasm, with the exported wrapper being a pointer to it.

warp/src/tesseract/wasm.rs
ashneverdawn: I managed to restructure it to satisfy what's expected by js for an asynciterator.

warp/src/tesseract/mod.rs
```diff
@@ -478,29 +499,96 @@ impl Tesseract {
...
-    /// To save to file using internal file path.
-    /// Note: Because we do not want to interrupt the functions due to it failing to
-    ///       save for whatever reason, this function will return `Result::Ok`
-    ///       regardless of success or error but would eventually log the error.
-    ///
-    /// TODO: Handle error without subjecting function to `Result::Err`
-    pub fn save(&self) -> Result<()> {
-        self.inner.save()
+/// Methods for wasm only
+#[cfg(target_arch = "wasm32")]
```
dariusc93, line 505:
```suggestion
```

warp/src/tesseract/mod.rs
```diff
@@ -478,29 +499,96 @@ impl Tesseract {
...
+        self.inner.save().map_err(|e| e.into())
+    }
+
+    #[wasm_bindgen(js_name = subscribe)]
+    pub fn subscribe_wasm(&self) -> AsyncIterator {
+        Into::<JsValue>::into(Subscription {
+            inner: self.inner.subscribe(),
+        })
+        .into()
+    }
```
dariusc93, line 550: I take it that these are renamed to prevent conflicts and to use the `JsError` vs splitting between wasm and non-wasm functions and only using `JsError` under `wasm` target?

warp/src/tesseract/mod.rs
```diff
@@ -478,29 +499,96 @@ impl Tesseract {
...
+/// Wraps in the TesseractEvent promise result in the js object expected by js async iterator
 #[cfg(target_arch = "wasm32")]
-impl Tesseract {
-    /// Used to load contents from local storage
-    pub fn load_from_storage(&self) -> Result<()> {
-        self.inner.load_from_storage()
+#[wasm_bindgen]
+struct TesseractEventPromiseResult {
+    pub value: TesseractEvent,
+    pub done: bool,
```
dariusc93, line 585: This does not look used or is this used when exported?

warp/Cargo.toml
```diff
@@ -58,17 +58,19 @@ paste.workspace = true
 tracing = { workspace = true }
 mediatype.workspace = true
 
+wasm-bindgen.workspace = true
+gloo.workspace = true
+web-sys.workspace = true
+js-sys.workspace = true
+console_error_panic_hook.workspace = true
+wasm-streams.workspace = true
+wasm-bindgen-futures.workspace = true
```
dariusc93, line 67: nit: Lets leave this under `[target.'cfg(target_arch = "wasm32")'.dependencies]` to minimize the dependencies at compile-time.

dariusc93 (commented):
Left a couple of comments but overall LGTM.

warp/src/tesseract/mod.rs
ashneverdawn: This is used by asynciterator. The Promise returned by `next()` must be a js `object` with the `value` and `done` fields.

warp/src/tesseract/mod.rs
ashneverdawn: correct. there are plenty of places that still need to be able to call these from inside the wasm and won't be able to use these js return types.

warp/src/tesseract/mod.rs
ashneverdawn: We do want to keep this here, or the compiler will complain on non-wasm that `wasm_bindgen` is not imported. It is only imported on wasm.

warp/Cargo.toml
ashneverdawn: done

warp/src/tesseract/wasm.rs
ashneverdawn: note however that as mentioned in the other comments, the subscription struct is able to be returned as an async iterator. The is only possible because it has the `next()` method with the expected signature and the `promise` returned by it returns a js `object` with `value` and `done` fields.

warp/src/tesseract/mod.rs
dariusc93: This shouldnt be needed since we have `#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]`, though we could probably drop this and use `#[cfg(target_arch = "wasm32")]`. Thoughts?

PR #510, opened 2024-05-03
export ipfs-identity-to-js

ashneverdawn:
**What this PR does** 📖

exports MultiPass to js
Includes an ipfs-identity from js example

#494 (partially)

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-ipfs/src/js_exports.rs
+++ b/extensions/warp-ipfs/src/js_exports.rs
@@ -100,7 +100,10 @@ fn to_identity_update_enum(
             Uint8Array::new(&value).to_vec(),
         )),
         IdentityUpdate::PicturePath => Ok(identity::IdentityUpdate::PicturePath(
-            value.as_string().ok_or(JsError::new("JsValue is not a string"))?.into(),
+            value
+                .as_string()
+                .ok_or(JsError::new("JsValue is not a string"))?
+                .into(),
         )),
         // IdentityUpdate::PictureStream => Ok(identity::IdentityUpdate::PictureStream(
         //     value.into()
@@ -110,7 +113,10 @@ fn to_identity_update_enum(
             Uint8Array::new(&value).to_vec(),
         )),
         IdentityUpdate::BannerPath => Ok(identity::IdentityUpdate::BannerPath(
-            value.as_string().ok_or(JsError::new("JsValue is not a string"))?.into(),
+            value
+                .as_string()
+                .ok_or(JsError::new("JsValue is not a string"))?
+                .into(),
         )),
         // IdentityUpdate::BannerStream => Ok(identity::IdentityUpdate::BannerStream(
         //     value.into()
--- a/extensions/warp-ipfs/src/js_exports.rs
+++ b/extensions/warp-ipfs/src/js_exports.rs
@@ -35,12 +35,12 @@ impl MultiPassBox {
         &self,
         id_variant: Identifier,
         id_value: JsValue,
-    ) -> Result<JsVecIdentity, JsError> {
+    ) -> Result<JsValue, JsError> {
         self.inner
             .get_identity(to_identifier_enum(id_variant, id_value)?)
             .await
             .map_err(|e| e.into())
-            .map(|ok| JsVecIdentity { inner: ok })
+            .map(|ok| serde_wasm_bindgen::to_value(&ok).unwrap())
     }
 
     pub async fn get_own_identity(&self) -> Result<Identity, JsError> {
@@ -56,19 +56,6 @@ impl MultiPassBox {
             .update_identity(to_identity_update_enum(option, value)?)
             .await
             .map_err(|e| e.into())
-    }
-}
-
-#[wasm_bindgen]
-pub struct JsVecIdentity {
-    inner: Vec<Identity>,
-}
-impl JsVecIdentity {
-    pub fn get(&self, i: usize) -> Identity {
-        self.inner[i].clone()
-    }
-    pub fn len(&self) -> usize {
-        self.inner.len()
     }
 }
 
--- a/extensions/warp-ipfs/src/js_exports.rs
+++ b/extensions/warp-ipfs/src/js_exports.rs
@@ -17,7 +17,7 @@ pub struct MultiPassBox {
 impl MultiPassBox {
     pub fn new(instance: WarpIpfs) -> Self {
         Self {
-            inner: Box::new(instance.clone()) as Box<_>,
+            inner: Box::new(instance) as Box<_>,
         }
     }
     pub async fn create_identity(
--- a/extensions/warp-ipfs/src/js_exports.rs
+++ b/extensions/warp-ipfs/src/js_exports.rs
@@ -1,7 +1,5 @@
-use std::str::FromStr;
-
-use crate::WarpIpfs;
 use js_sys::Uint8Array;
+use std::str::FromStr;
 use warp::crypto::DID;
 use warp::multipass::{
     identity::{self, Identity, IdentityProfile},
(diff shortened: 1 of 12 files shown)


extensions/warp-ipfs/src/js_exports.rs
```diff
@@ -0,0 +1,180 @@
...
+use js_sys::Uint8Array;
+use warp::crypto::DID;
+use warp::multipass::{
+    identity::{self, Identity, IdentityProfile},
+    MultiPass,
+};
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub struct MultiPassBox {
```
dariusc93, line 13: The name may not be suitable here. Not sure what the proper name should be though.

dariusc93: I wonder if it might be better to create a macro rule so we could pass `WarpIpfs` (or any struct or item that implements `MultiPass` and its traits) so we could do something like `generate_wam!(WarpIpfs);`  when the target is `wasm32-unknown-unknown` it would generate the functions accordingly, maybe even a constructor function that could be used from js?

extensions/warp-ipfs/src/js_exports.rs
```diff
@@ -0,0 +1,180 @@
...
+    inner: Box<dyn MultiPass>,
+}
+#[wasm_bindgen]
+impl MultiPassBox {
+    pub fn new(instance: WarpIpfs) -> Self {
+        Self {
+            inner: Box::new(instance.clone()) as Box<_>,
```
dariusc93, line 20: Clone looks redundant

```suggestion
            inner: Box::new(instance) as Box<_>,
```

extensions/warp-ipfs/src/js_exports.rs
```diff
@@ -0,0 +1,180 @@
...
+            .create_identity(username.as_deref(), passphrase.as_deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    pub async fn get_identity(
+        &self,
+        id_variant: Identifier,
+        id_value: JsValue,
+    ) -> Result<JsVecIdentity, JsError> {
```
dariusc93, line 38: We could turn the `Vec<Identity>` into `JsValue` with `serde-wasm-bindgen`, could we not?

warp/src/multipass/identity.rs
```diff
@@ -61,11 +63,26 @@ impl IdentityProfile {
...
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = set_identity))]
+    pub fn set_identity_wasm(&mut self, identity: Identity) {
+        self.identity = identity;
+    }
+
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = passphrase))]
+    pub fn passphrase_wasm(&self) -> Option<String> {
+        self.passphrase.as_ref().map(|phrase| phrase.to_string())
+    }
+}
```
dariusc93, line 80: here, what we could do instead is just split the impl between wasm32 and non-wasm32 so we dont have to worry about the naming of the function through wasm-bindgen

warp/src/multipass/identity.rs
```diff
@@ -207,23 +225,49 @@ pub struct Identity {
...
+    pub fn set_created_wasm(&mut self, time: js_sys::Date) {
+        self.created = time.into();
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = set_modified))]
+    pub fn set_modified_wasm(&mut self, time: js_sys::Date) {
+        self.modified = time.into();
+    }
+}
```
dariusc93, line 262: ditto

warp/src/multipass/identity.rs
```diff
@@ -233,11 +277,41 @@ impl Identity {
...
+    pub fn created_wasm(&self) -> js_sys::Date {
+        self.created.into()
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = modified))]
+    pub fn modified_wasm(&self) -> js_sys::Date {
+        self.modified.into()
+    }
+}
```
dariusc93, line 313: :+1:

warp/src/multipass/identity.rs
```diff
@@ -207,23 +225,49 @@ pub struct Identity {
...
         self.short_id = id.into()
     }
 
     pub fn set_did_key(&mut self, pubkey: DID) {
         self.did_key = pubkey
     }
-
-    pub fn set_status_message(&mut self, message: Option<String>) {
-        self.status_message = message
-    }
```
dariusc93, line 225: Why is this function moved behind wasm cfg attr?

warp/src/multipass/identity.rs
```diff
@@ -246,10 +320,6 @@ impl Identity {
         self.did_key.clone()
     }
 
-    pub fn status_message(&self) -> Option<String> {
-        self.status_message.clone()
-    }
```
dariusc93, line 251: ditto

extensions/warp-ipfs/src/js_exports.rs
ashneverdawn: nice catch. done

warp/src/multipass/identity.rs
ashneverdawn: Thanks, I split it now, however note that the function still can't have the original name. On wasm the altered signature is there for being callable from js. But we still need the original signature when it is called internally within the wasm by other parts of our code. Since we can't have 2 methods with the same name, we need to pick a different name for the one with the altered function signature, hence the `_wasm`. The wasm-bindgen renaming is only there to get rid of the `_wasm` on the js side since it is redundant.

warp/src/multipass/identity.rs
ashneverdawn: see above comment

warp/src/multipass/identity.rs
ashneverdawn: see above comment

warp/src/multipass/identity.rs
ashneverdawn: see above comment

extensions/warp-ipfs/src/js_exports.rs
ashneverdawn: nice. didn't know about `serde-wasm-bindgen`. I made the change now. Also, worth mentioning `wasm-bindgen` has recently merged in support for `Vec<T>` in async functions but have not yet released a new version. When the next version is out we won't need `serde-wasm-bindgen` anymore for this particular case.

warp/src/multipass/identity.rs
dariusc93: As I think about it, I overlooked that I was borrowing the `passphrase` field in `IdentityProfile::passphrase`.

warp/src/multipass/identity.rs
dariusc93: Hmm, thought i saw a `#[cfg(target_arch = "wasm32")]` as well as `#[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen)]` hence the question.

extensions/warp-ipfs/src/js_exports.rs
ashneverdawn: As discussed, a new `WarpIpfs` returns a `WarpInstance` which contains all the boxed traits.

dariusc93 (changes requested):
Lets move `js_exports` to `warp` crate but besides that LGTM.

dariusc93 (approved):

PR #519, opened 2024-05-17
chore: export raygun to js

ashneverdawn:
**What this PR does** 📖

Exposes Raygun to js

Relating to: #494

**Additional comments** 🎤


warp/src/js_exports/raygun.rs
```diff
@@ -11,3 +23,746 @@ impl RayGunBox {
...
+    pub fn limit(&self) -> Option<u8> {
+        self.inner.limit()
+    }
+
+    pub fn skip(&self) -> Option<i64> {
+        self.inner.skip()
+    }
+
+    pub fn keyword(&self) -> Option<String> {
+        self.inner.keyword().clone()
```
dariusc93, line 541:
```suggestion
        self.inner.keyword()
```

warp/src/js_exports/raygun.rs
```diff
@@ -11,3 +23,746 @@ impl RayGunBox {
...
+            .set_messages_type(serde_wasm_bindgen::from_value(ty).unwrap());
+    }
+
+    pub fn date_range(&self) -> JsValue {
+        serde_wasm_bindgen::to_value(&self.inner.date_range()).unwrap()
+    }
+
+    pub fn range(&self) -> JsValue {
+        serde_wasm_bindgen::to_value(&self.inner.range()).unwrap()
+    }
```
dariusc93, line 530: Would the range produce a value `JsValue`?

warp/src/js_exports/raygun.rs
```diff
@@ -11,3 +23,746 @@ impl RayGunBox {
...
+        Self {
+            inner: raygun::MessageOptions::default(),
+        }
+    }
+
+    pub fn set_date_range(&mut self, range: JsValue) {
+        self.inner = self
+            .inner
+            .clone()
+            .set_date_range(serde_wasm_bindgen::from_value(range).unwrap());
```
dariusc93, line 479: wouldnt it have been better to mutate `inner` directly instead of cloning inner and mutating that instance?

dariusc93 (commented):
Left some comments but overall LGTM. 

I will say though that we can skip on many of the setters as I dont really imagine a use-case for any frontend would need to set most of them.

warp/src/js_exports/raygun.rs
ashneverdawn: I would have preferred that, but the way `MessageOptions` is implemented, all of its methods move the instance, which is useful for chaining methods, but doesn't really play nice in this case since the data lives on the wasm side and js only holds a reference to it. 

https://github.com/Satellite-im/Warp/blob/bc7a794302092987816b05e99857a1fefeb6d1dc/warp/src/raygun/mod.rs#L133-L140

I do think it would be nice to refactor this is another PR, but I opted for cloning here at the moment to reduce the amount of code to write/change in this PR compared to other alternatives:

Option 1 - Change `MessageOptions` to act on a reference of self instead, and updating all the places throughout the code base where these methods are called. (I prefer this for consistency with the rest of the code)

Option 2 - Create a new struct similar to MessageOptions with to and from conversion functions, using only js compatible types to be able to have it live in js memory and chain method calls from there.

warp/src/js_exports/raygun.rs
ashneverdawn: Unfortunately it does not.

warp/src/js_exports/raygun.rs
dariusc93: We could make that change in a future PR

warp/src/js_exports/raygun.rs
dariusc93: hmm ok. For now we can omit these from being exported since I dont believe access to those values in JS would be needed.

warp/src/js_exports/raygun.rs
ashneverdawn: done

ashneverdawn:
> Left some comments but overall LGTM.
> 
> I will say though that we can skip on many of the setters as I dont really imagine a use-case for any frontend would need to set most of them.

Thanks. I removed all the unnecessary setters. Looks like `MessageOptions` is the only one that needed them.

warp/src/js_exports/raygun.rs
ashneverdawn: To clarify, `serde_wasm_bindgen::to_value()` does turn `Range<...>` into a JsValue. 
I just meant that it wouldn't otherwise.

ashneverdawn: Yeah, in fact I'm noticing this is part of MessageOptions which is only used for input from js side. So I will get rid of all the getters, since it only needs setters.

ashneverdawn: done 👍

dariusc93 (approved):

PR #532, opened 2024-06-04
chore: export constellation put_stream to js

ashneverdawn:
**What this PR does** 📖

Exports constellation's put_stream to js. 
Adds an example usage in js

Related to #494 

**Additional comments** 🎤


warp/src/constellation/mod.rs
```diff
@@ -47,7 +47,7 @@ impl Stream for ConstellationEventStream {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, serde::Serialize)]
```
dariusc93, line 50: Lets put this derive behind a cfg_attr for wasm32.

dariusc93 (commented):
Left one suggestion but overall LGTM.

warp/src/constellation/mod.rs
ashneverdawn: as you later suggested on discord, I instead made a copy of this enum, that way it can also hold the error information instead of skipping it on serialization. (Error isn't serializable) It turns out JsError wasn't serializable either, so I used String.

dariusc93 (approved):

PR #536, opened 2024-06-07
chore: export missing js functions

ashneverdawn:
**What this PR does** 📖

Added functions that were missing for js.
Reordered js functions to be in the same order as in the trait so that it's easier to see if any are missing.

Related to: #494

**Additional comments** 🎤

### DIFF
--- a/warp/src/js_exports/constellation.rs
+++ b/warp/src/js_exports/constellation.rs
@@ -30,8 +30,42 @@ impl ConstellationBox {
         }
     }
 
+    pub fn current_size(&self) -> usize {
+        self.inner.current_size()
+    }
+
     pub fn max_size(&self) -> usize {
         self.inner.max_size()
+    }
+
+    pub fn select(&mut self, path: &str) -> Result<(), JsError> {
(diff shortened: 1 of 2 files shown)


warp/src/js_exports/constellation.rs
```diff
@@ -30,10 +30,44 @@ impl ConstellationBox {
...
+    pub fn select(&mut self, path: &str) -> Result<(), JsError> {
+        self.inner.select(path).map_err(|e| e.into())
+    }
+
+    pub fn set_path(&mut self, path: String) {
+        self.inner.set_path(path.into())
+    }
+
+    pub fn get_path(&self) -> String {
+        self.inner.get_path().to_str().unwrap().into()
```
dariusc93, line 50: We could use `PathBuf::to_string_lossy` here instead if permitted by wasm32 target.

dariusc93 (approved):
Left one small comment but LGTM

warp/src/js_exports/constellation.rs
ashneverdawn: thanks. done.

PR #541, opened 2024-06-11
wasm: add attachment and group raygun methods

Flemmli97:
**What this PR does** 📖

Adds the missing stuff for raygun in regards to attachment and groups


warp/src/js_exports/raygun.rs
```diff
@@ -301,6 +304,137 @@ impl RayGunBox {
...
+    /// Add a recipient to the conversation
+    pub async fn add_recipient(
+        &mut self,
+        conversation_id: String,
+        recipient: String,
+    ) -> Result<(), JsError> {
+        self.inner
+            .add_recipient(
+                Uuid::from_str(&conversation_id).unwrap(),
+                &DID::from_str(&recipient).unwrap_or_default(),
```
dariusc93, line 331: You would want to avoid using `unwrap_or_default` here as if the key is invalid, it would generate a new key which would not be targeted towards the specific identity. 
```suggestion
                &DID::from_str(&recipient).unwrap(),
```

warp/src/js_exports/raygun.rs
```diff
@@ -301,6 +304,137 @@ impl RayGunBox {
...
+    /// Remove a recipient from the conversation
+    pub async fn remove_recipient(
+        &mut self,
+        conversation_id: String,
+        recipient: String,
+    ) -> Result<(), JsError> {
+        self.inner
+            .remove_recipient(
+                Uuid::from_str(&conversation_id).unwrap(),
+                &DID::from_str(&recipient).unwrap_or_default(),
```
dariusc93, line 346:
```suggestion
                &DID::from_str(&recipient).unwrap(),
```

warp/src/js_exports/raygun.rs
```diff
@@ -301,6 +304,137 @@ impl RayGunBox {
...
+                PathBuf::from(path),
+            )
+            .await
+            .map_err(|e| e.into())
+            .map(|ok| {
+                AsyncIterator::new(Box::pin(
+                    ok.map(|s| serde_wasm_bindgen::to_value(&Progression::from(s)).unwrap()),
+                ))
+            })
+    }
```
dariusc93, line 408: We would need to omit/exclude this since this function is specific towards downloading to a specific path on the filesystem and we dont have that type of access on the browser. Instead, use `RayGun::download_stream`.

warp/src/js_exports/raygun.rs
```diff
@@ -301,6 +304,137 @@ impl RayGunBox {
...
+            )
+            .await
+            .map_err(|e| e.into())
+            .map(|(id, ok)| AttachmentResult {
+                file: id.to_string(),
+                stream: AsyncIterator::new(Box::pin(
+                    ok.map(|s| serde_wasm_bindgen::to_value(&AttachmentKind::from(s)).unwrap()),
+                )),
+            })
+    }
```
dariusc93, line 383: As a note, you would not be able to use `Location::Path` since that relies on the filesystem. I would suggest uploading to constellation first and use `Location::Constellation` for the time being until `Location` supports piping a stream for a specific file.

dariusc93 (changes requested):

warp/src/js_exports/raygun.rs
dariusc93: You can but that information isnt passable here and its more of a security thing for the browser.

warp/src/js_exports/raygun.rs
dariusc93: Since we cant access the filesystem in wasm, we would need to rely on other means since we cannot read a file using `Location::Path` from the browser, which in this case would use contellation or passing a valid stream of bytes that we can poll and pass into constellation. A workaround for now would be to use `Location::Constellation`, which would require the frontend to upload to constellation first before attaching it in a message since it would already be available there.

Flemmli97: i added the version that now uses a stream

dariusc93: Lets focus on the streams in a separate PR as we can use reuse `RayGun::attach` by adding streams to `Location` instead of introducing a new object and new function

Flemmli97: i added it as something new cause i think streams are not cloneable? which location requires

dariusc93: We can have something inclusive to `js_export` so it could probably either be split (similar to what is done for `MultiPass::get_identity`) or have a separate function exclusive to js export

dariusc93 (commented):
With #544 merged, we could do something similar to what we do in multipass for `MultiPass::get_identity` variant but also factoring in how we do streams similar to `Constellation::put_stream` variant  https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/multipass.rs#L41-L51

https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/constellation.rs#L86-L103

https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/stream.rs#L49-L76

CC @ashneverdawn

ashneverdawn:
> With #544 merged, we could do something similar to what we do in multipass for `MultiPass::get_identity` variant but also factoring in how we do streams similar to `Constellation::put_stream` variant
> 
> https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/multipass.rs#L41-L51
> 
> https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/constellation.rs#L86-L103
> 
> https://github.com/Satellite-im/Warp/blob/d8bb96e1be4b5bf9d53cd3551958e89299b78957/warp/src/js_exports/stream.rs#L49-L76
> 
> CC @ashneverdawn

Agreed 👍

warp/src/js_exports/raygun.rs
```diff
@@ -643,3 +748,66 @@ impl Message {
         self.inner.replied().map(|uuid| uuid.to_string())
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[wasm_bindgen]
+pub struct AttachmentFile {
+    name: String,
+    stream: web_sys::ReadableStream,
+}
```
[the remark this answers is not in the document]
Flemmli97: added that

ashneverdawn (approved):

dariusc93 (approved):

PR #550, opened 2024-06-17
chore: Remove unused function

ashneverdawn:
**What this PR does** 📖

Uses the const `MAX_STATUS_LENGTH` instead of `512`. It was already being used everywhere else in `warp-ipfs`, but not in this case. Also, `512` is value of `MAX_STATUS_LENGTH`

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-ipfs/src/store/identity.rs
+++ b/extensions/warp-ipfs/src/store/identity.rs
@@ -39,7 +39,7 @@ use warp::{
 
 use crate::{
     config::{self, Discovery as DiscoveryConfig},
-    store::{discovery::Discovery, topics::PeerTopic, DidExt, PeerIdExt},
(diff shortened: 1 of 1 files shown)


dariusc93 (approved):

dariusc93:
Can you correct the errors before merging?

ashneverdawn:
> Can you correct the errors before merging?
> 
> EDIT: Looks like its a function that isnt used. Might be best to remove that instead?

Somehow I left the constant out from the use statement, my mistake. But yeah, if it isn't used by anything in Warp or Uplink, then makes sense to remove it.

ashneverdawn:
> Can you correct the errors before merging?
> 
> EDIT: Looks like its a function that isnt used. Might be best to remove that instead?

Done.

dariusc93 (approved):

PR #555, opened 2024-06-27
feat(raygun): Add size to download stream

Flemmli97:
**What this PR does** 📖

Appends the file size to download streams in raygun

### DIFF
--- a/extensions/warp-ipfs/src/lib.rs
+++ b/extensions/warp-ipfs/src/lib.rs
@@ -1658,7 +1658,7 @@ impl RayGunAttachment for WarpIpfs {
         conversation_id: Uuid,
         message_id: Uuid,
         file: &str,
-    ) -> Result<BoxStream<'static, Result<Vec<u8>, Error>>, Error> {
+    ) -> Result<(usize, BoxStream<'static, Result<Vec<u8>, Error>>), Error> {
         self.messaging_store()?
             .download_stream(conversation_id, message_id, file)
             .await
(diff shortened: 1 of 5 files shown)


extensions/warp-ipfs/src/lib.rs
```diff
@@ -1658,7 +1658,7 @@ impl RayGunAttachment for WarpIpfs {
         conversation_id: Uuid,
         message_id: Uuid,
         file: &str,
-    ) -> Result<BoxStream<'static, Result<Vec<u8>, Error>>, Error> {
+    ) -> Result<(usize, BoxStream<'static, Result<Vec<u8>, Error>>), Error> {
```
dariusc93, line 1661: Instead of making it a tuple, we should probably create a struct that is a stream that would house that metadata while still serving the bytes itself.

dariusc93 (commented):
Im not to sure on if this should be needed since you could get the size of the file from scanning the attachments from the `Message` itself.

Flemmli97:
> Im not to sure on if this should be needed since you could get the size of the file from scanning the attachments from the `Message` itself.

i guess thats also true but since the size is already fetched but not sent i would say to just also sent it with the data instead of needing to fetch again

extensions/warp-ipfs/src/lib.rs
Flemmli97: ye i can change that

dariusc93:
> i guess thats also true but since the size is already fetched but not sent i would say to just also sent it with the data instead of needing to fetch again

The thing is that you would have to access the attachments in the message anyway, in which case would include `File::size` that represents the size of the attached file so adding it to the function may actually be redundant.

Flemmli97:
actually you are right. the ui already has information about the size. im closing this

PR #556, opened 2024-06-27
feature(raygun): add filesize to message attach

Flemmli97:
**What this PR does** 📖

Adds a size field to `Location::Stream`. And changes the wasm part to also pass in the size with the stream

### DIFF
--- a/warp/src/js_exports/raygun.rs
+++ b/warp/src/js_exports/raygun.rs
@@ -6,7 +6,7 @@ use crate::{
         PinState, RayGun, ReactionState,
     },
 };
-use futures::StreamExt;
+use futures::{StreamExt};
 use js_sys::Promise;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


warp/src/js_exports/raygun.rs
```diff
@@ -6,7 +6,7 @@ use crate::{
         PinState, RayGun, ReactionState,
     },
 };
-use futures::StreamExt;
+use futures::{StreamExt};
```
dariusc93, line 9:
```suggestion
use futures::StreamExt;
```

warp/src/js_exports/raygun.rs
```diff
@@ -753,24 +753,37 @@ impl Message {
...
-    pub fn new(file: String, stream: Option<web_sys::ReadableStream>) -> Self {
+    pub fn new(file: String, stream: Option<AttachmentStream>) -> Self {
         Self { file, stream }
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[wasm_bindgen]
+pub struct AttachmentStream {
+    size: usize,
```
dariusc93, line 769: We should probably make this type into `Option<usize>` and pass that option through to `Location::Stream`

dariusc93 (changes requested):

dariusc93:
Can you also update the example?

PR #564, opened 2024-07-11
add persistent chat example for js

ashneverdawn:
**What this PR does** 📖

adds a js example for conversation with persistence enabled (minimal_basic). (I used this to help debug conversation desync issue on browser refresh.)

**Additional comments** 🎤


ashneverdawn:
> LGTM. Maybe we should add a note that the persistent side require indexeddb and/or localstorage and that those storages should be cleared if they wish for things to be refreshed

Good point 👍  I took the opportunity to add descriptions to all js examples

dariusc93 (approved):

PR #569, opened 2024-07-18
feat(wasm): Add more stricter raygun typing

Flemmli97:
**What this PR does** 📖

- Adds more stricter typing to `list_conversation` and `fetch_messages`:
  - `list_conversation` returns now a `ConversationList` which holds an array of `raygun::Conversation`. `convs` then maps and returns the conversations as a wasm `Conversation`.
  - There is not really a way to do this for `fetch_messages` as that returns multiple types of data. Instead i added a static function `message_from` that can turn a Json Object of `raygun::Message` into a wasm `Message` so something like following is possible:
Screenshot 2024-07-18 at 19 13 23

### DIFF
--- a/warp/src/js_exports/raygun.rs
+++ b/warp/src/js_exports/raygun.rs
@@ -76,19 +76,12 @@ impl RayGunBox {
     }
 
     /// List all active conversations
-    pub async fn list_conversations(&self) -> Result<JsValue, JsError> {
+    pub async fn list_conversations(&self) -> Result<ConversationList, JsError> {
         self.inner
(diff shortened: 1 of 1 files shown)


ashneverdawn (approved):
Nice! I tested it and it works as expected

ashneverdawn:
As a note for future reference, the next release of `wasm-bindgen` adds support for returning `Vec` from async functions, so once it's out, we'll be able to simply return a `Vec` without needing `serde_wasmbindgen`

dariusc93 (approved):

PR #585, opened 2024-08-08
refactor: Move message, reactions and attachment into a `MessageDocument`

dariusc93:
**What this PR does** 📖
- moves the content of messages into MessageDocument instead of storing it as a dag
- moves attachments, and reactions into `MessageDocument`. 

- Previous messages prior to this PR will not be compatible.

**Additional comments** 🎤
- This change allow us to reduce the amount of resolving needed for the message document since previously we would store the encrypted text, attachments, and reactions into its own block to keep the document size small when sending it over pubsub or storing it, however since the message document has other metadata that is also small (ie `FileAttachment`), it might be a bit redundant to keep the data itself in its own block. Moving it to the upper level of the message document would allow it to be available right away while taking advantage of the small document and enforced limit checks to keep the overall document small.
- We use `Bytes` instead of `Vec<u8>` to reduce allocation of the bytes of messages when cloning, although this *might* change in the future.


dariusc93:
Side Note:
- Apart of this, we could probably migrate some parts out of its own document and into the message document itself and enforce a size limit there, but may be done in a separate refactoring PR

ashneverdawn (approved):

PR #588, opened 2024-08-12
feat: log to console wasm

Flemmli97:
**What this PR does** 📖

Adds option to print the warp logs to console on wasm. Requires calling the exported `trace` function so this is not on by default

### DIFF
--- a/warp/src/js_exports/mod.rs
+++ b/warp/src/js_exports/mod.rs
@@ -19,6 +19,11 @@ pub fn initialize() {
 }
 
 #[wasm_bindgen]
+pub fn trace() {
+    tracing_wasm::set_as_global_default();
+}
+
+#[wasm_bindgen]
 pub struct WarpInstance {
     multipass: MultiPassBox,
     raygun: RayGunBox,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,6 +107,7 @@ wasm-streams = "0.4"
 wasm-bindgen-futures = "0.4"
 serde-wasm-bindgen = "0.4"
 send_wrapper = "0.6.0"
+tracing-wasm = "0.2.0"
 
 # Blink related crates
 # av-data is needed to use libaom. need to ensure that Warp and libaom use the same version of av-data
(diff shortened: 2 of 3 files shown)


warp/src/js_exports/mod.rs
```diff
@@ -18,6 +18,11 @@ pub fn initialize() {
     std::panic::set_hook(Box::new(console_error_panic_hook::hook));
 }
 
+#[wasm_bindgen]
+pub fn trace() {
+    tracing_wasm::set_as_global_default();
```
dariusc93, line 23: I think we should probably use the configuration and maybe create an enum for setting the log level. Thoughts?

cc @ashneverdawn

ashneverdawn: Sure, that would work. However, maybe we should consider having 2 separate wasm builds of warp? `warp-wasm` for prod with everything stripped out for optimal performance, and another called `warp-wasm-dev` using compile time features flags to enable dev options like enabling tracing & setting log level. And we'd switch between `warp-wasm` and `warp-wasm-dev` in `package.json`.

I say this is because wasm size can easily become the main bottleneck for performance and initial load times when a lot of bloat gets included - and it only gets worse on lower bandwidth and/or mobile devices. 

Don't know if we're at the point where it makes a big difference yet though. It'd be interesting to compare the current difference in wasm size with or without tracing included in the build. If the difference is negligible, then what I'm proposing might be more trouble than it's worth at this point.

tooshel: why would you need a completely seperate repo? you can't change that with an env var and "npm install"?

dariusc93: > why would you need a completely seperate repo? you can't change that with an env var and "npm install"?

So the build could be done there i.e via CI (at least for starters), but for the packages itself, I would not know for sure how npm handles that these days. @ashneverdawn or @Flemmli97 may be able to answer that.

dariusc93 (approved):

ashneverdawn (approved):

PR #590, opened 2024-08-13
fix(wasm): fix attachments having no thumbnail

Flemmli97:
**What this PR does** 📖

Fixes the problem of attachment thumbnails missing by writing each file to js individually

### DIFF
--- a/warp/src/js_exports/raygun.rs
+++ b/warp/src/js_exports/raygun.rs
@@ -748,7 +748,7 @@ impl Message {
         self.inner
             .attachments()
             .iter()
-            .map(|v| serde_wasm_bindgen::to_value(v).unwrap())
+            .filter_map(|v| serde_wasm_bindgen::to_value(v).ok())
             .collect()
     }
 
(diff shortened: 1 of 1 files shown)


warp/src/js_exports/raygun.rs
```diff
@@ -744,8 +744,12 @@ impl Message {
         self.inner.lines()
     }
 
-    pub fn attachments(&self) -> JsValue {
-        serde_wasm_bindgen::to_value(&self.inner.attachments()).unwrap()
+    pub fn attachments(&self) -> Vec<JsValue> {
+        self.inner
+            .attachments()
+            .iter()
+            .map(|v| serde_wasm_bindgen::to_value(v).unwrap())
```
dariusc93, line 751: Since a PR is already up to reduce unwrapping, where possible, we should follow the same scheme here too.

```suggestion
            .filter_map(|v| serde_wasm_bindgen::to_value(v).ok())
```

dariusc93 (changes requested):

dariusc93 (approved):

PR #591, opened 2024-08-14
feat(wasm): expose some config functions to wasm

Flemmli97:
**What this PR does** 📖

Exposes some functions to allow modifying the config during init

### DIFF
--- a/extensions/warp-ipfs/src/config.rs
+++ b/extensions/warp-ipfs/src/config.rs
@@ -285,43 +285,43 @@ impl Config {
 
 #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen)]
 impl Config {
-    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = path))]
-    pub fn path_mut_wasm(&mut self, path: String) {
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen)]
+    pub fn with_path(&mut self, path: String) {
         self.path = Some(PathBuf::from(&path))
     }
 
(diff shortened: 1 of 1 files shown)


extensions/warp-ipfs/src/config.rs
```diff
@@ -283,6 +283,49 @@ impl Config {
...
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = set_thumbnail_size))]
+    pub fn thumbnail_size_mut_wasm(&mut self, size_x: u32, size_y: u32) {
+        self.thumbnail_size = (size_x, size_y)
+    }
+
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen::prelude::wasm_bindgen(js_name = thumbnail_exact_format))]
+    pub fn thumbnail_exact_format_mut_wasm(&mut self, exact: bool) {
+        self.thumbnail_exact_format = exact
+    }
+}
```
dariusc93, line 327: instead of including `_mut` we could just have most of these functions use `set` instead. Eg `set_thumbnail_size`, `set_thumbnail_exact_format`, etc.

dariusc93 (changes requested):
Left a small comment but lgtm.

extensions/warp-ipfs/src/config.rs
Flemmli97: i used mut to reflect them being the equivalent wasm function as the rust ones are called mut. but i guess changing it is fine

dariusc93: The mut part was related to the mutable reference being returned in those functions. That naming scheme probably wouldnt be needed when exporting out.

dariusc93 (approved):

PR #600, opened 2024-08-29
fix(constellation): remove target directory after recursive removal

Flemmli97:
**What this PR does** 📖

Fixes recursive deletion not deleting the folder itself

### DIFF
--- a/extensions/warp-ipfs/src/store/files.rs
+++ b/extensions/warp-ipfs/src/store/files.rs
@@ -34,7 +34,8 @@ use crate::{
 };
 
 use super::{
-    document::root::RootDocumentMap, event_subscription::EventSubscription,
+    document::root::{self, RootDocumentMap},
+    event_subscription::EventSubscription,
     MAX_THUMBNAIL_STREAM_SIZE,
 };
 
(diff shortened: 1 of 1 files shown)


extensions/warp-ipfs/src/store/files.rs
```diff
@@ -1275,6 +1275,9 @@ async fn _remove(ipfs: &Ipfs, root: &Directory, item: &Item) -> Result<(), Error
                     }
                 }
             }
+            if let Err(e) = root.remove_item(&directory.name()) {
+                tracing::error!(error = %e, item_name = %directory.name(), "unable to remove directory");
+            }
```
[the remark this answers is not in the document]
dariusc93: It would be best in case something is inserted into that directory while we are scanning through and removing the items. A simple size check should do since empty directories dont count towards the size.

extensions/warp-ipfs/src/store/files.rs
```diff
@@ -1275,6 +1276,11 @@ async fn _remove(ipfs: &Ipfs, root: &Directory, item: &Item) -> Result<(), Error
                     }
                 }
             }
+            if directory.size() != 0 {
```
dariusc93, line 1279:
```suggestion
            if directory.size() == 0 {
```

extensions/warp-ipfs/src/store/files.rs
```diff
@@ -34,7 +34,8 @@ use crate::{
 };
 
 use super::{
-    document::root::RootDocumentMap, event_subscription::EventSubscription,
+    document::root::{self, RootDocumentMap},
```
dariusc93, line 37: Causing linting to fail

```suggestion
    document::root::RootDocumentMap,
```

dariusc93 (approved):

PR #603, opened 2024-09-06
chore: move js exports to warp wasm repo

### LINKED ISSUE #581: chore(wasm): Move exported bindings to its own repo
After talking to @ashneverdawn, I do feel that the js exported functions should be move to [warp-wasm](https://github.com/Satellite-im/warp-wasm) repo, which would reduce breaking changes and provide more clarity on updates, while reducing the repeated code in the warp repo and allow direct changes to the exported code if needed without needing to update the warp then update the warp-wasm repo. Given that `warp` and `warp-ipfs` itself can compile to `wasm32-unknown-unknown`, this would likely be the best move.

ashneverdawn:
**What this PR does** 📖

- Removes `wasm_bindgen` from types and functions exported to js. This is now done from the warp-wasm repo instead.
- Once this is merged, there will be a commit hash on Warp `main` that I can use for warp-wasm `move-js-exports-to-warp-wasm-repo` branch's Warp dependency. 

Resolves #581


warp/src/tesseract/mod.rs
```diff
@@ -534,52 +525,11 @@ impl Tesseract {
...
-    }
-
-    #[wasm_bindgen(js_name = subscribe)]
-    pub fn subscribe_wasm(&self) -> AsyncIterator {
-        AsyncIterator::new(Box::pin(self.subscribe().map(|t| Into::<JsValue>::into(t))))
-    }
-
     /// Used to load contents from local storage
-    pub fn load_from_storage(&self) -> std::result::Result<(), JsError> {
+    pub fn load_from_storage(&self) -> std::result::Result<(), Error> {
```
dariusc93, line 530: I wonder if we should move this function in whole to `warp-wasm` repo. Thoughts?

dariusc93 (commented):
LGTM. Left one comment and to add an additional question: Should we hold off on merging this until the move is complete on the other repo?

warp/src/tesseract/mod.rs
ashneverdawn: Would have been nice, but `load_from_storage` accesses private stuff from InnerTesseract, so we wouldn't be able to copy it out as is. But perhaps in a future PR we could rewrite it to achieve the same behavior via other public exported methods?

PR #604, opened 2024-09-12
chore: refactor group settings for roles and permissions

ashneverdawn:
**What this PR does** 📖

- Refactors GroupSettings for roles and permissions

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-ipfs/src/store/conversation.rs
+++ b/extensions/warp-ipfs/src/store/conversation.rs
@@ -25,6 +25,7 @@ use warp::{
     crypto::{cipher::Cipher, hash::sha256_iter, DIDKey, Ed25519KeyPair, KeyMaterial, DID},
     error::Error,
     raygun::{
+        Permission,
         Conversation, ConversationSettings, ConversationType, DirectConversationSettings,
         GroupSettings, Message, MessageOptions, MessagePage, MessageReference, MessageType,
         Messages, MessagesType,
@@ -257,14 +258,14 @@ impl ConversationDocument {
 
 impl ConversationDocument {
     pub fn sign(&mut self, keypair: &Keypair) -> Result<(), Error> {
-        if let ConversationSettings::Group(settings) = self.settings {
+        if let ConversationSettings::Group(settings) = &self.settings {
             assert_eq!(self.conversation_type(), ConversationType::Group);
             let did = keypair.to_did()?;
             let Some(creator) = self.creator.clone() else {
                 return Err(Error::PublicKeyInvalid);
             };
 
-            if !settings.members_can_add_participants() && !creator.eq(&did) {
+            if !settings.user_has_permission(&did, Permission::AddParticipantsToGroup) && !creator.eq(&did) {
                 return Err(Error::PublicKeyInvalid);
             }
 
@@ -275,7 +276,7 @@ impl ConversationDocument {
             let construct = warp::crypto::hash::sha256_iter(
                 [
                     Some(self.id().into_bytes().to_vec()),
-                    (!settings.members_can_change_name())
+                    (!settings.user_has_permission(&did, Permission::ChangeGroupName))
                         .then(|| self.name.as_deref().map(|s| s.as_bytes().to_vec()))
                         .flatten(),
                     self.description.as_ref().map(|d| d.as_bytes().to_vec()),
@@ -287,7 +288,7 @@ impl ConversationDocument {
                             .iter()
                             .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
                     )),
-                    (!settings.members_can_add_participants()).then_some(Vec::from_iter(
+                    (!settings.user_has_permission(&did, Permission::AddParticipantsToGroup)).then_some(Vec::from_iter(
                         self.recipients
                             .iter()
                             .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
@@ -304,13 +305,14 @@ impl ConversationDocument {
     }
 
     pub fn verify(&self) -> Result<(), Error> {
-        if let ConversationSettings::Group(settings) = self.settings {
+        if let ConversationSettings::Group(settings) = &self.settings {
             assert_eq!(self.conversation_type(), ConversationType::Group);
             let Some(creator) = &self.creator else {
                 return Err(Error::PublicKeyInvalid);
             };
 
             let creator_pk = creator.to_public_key()?;
+            let did = creator_pk.to_did()?;
 
             let Some(signature) = &self.signature else {
                 return Err(Error::InvalidSignature);
@@ -341,7 +343,7 @@ impl ConversationDocument {
                                 .iter()
                                 .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
                         )),
-                        (!settings.members_can_add_participants()).then_some(Vec::from_iter(
+                        (!settings.user_has_permission(&did, Permission::AddParticipantsToGroup)).then_some(Vec::from_iter(
                             self.recipients
                                 .iter()
                                 .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
@@ -353,7 +355,7 @@ impl ConversationDocument {
                 ConversationVersion::V2 => warp::crypto::hash::sha256_iter(
                     [
                         Some(self.id().into_bytes().to_vec()),
-                        (!settings.members_can_change_name())
+                        (!settings.user_has_permission(&did, Permission::ChangeGroupName))
                             .then(|| self.name.as_deref().map(|s| s.as_bytes().to_vec()))
                             .flatten(),
                         Some(creator.to_string().as_bytes().to_vec()),
@@ -364,7 +366,7 @@ impl ConversationDocument {
                         )),
                         self.icon.map(|s| s.hash().digest().to_vec()),
                         self.banner.map(|s| s.hash().digest().to_vec()),
-                        (!settings.members_can_add_participants()).then_some(Vec::from_iter(
+                        (!settings.user_has_permission(&did, Permission::AddParticipantsToGroup)).then_some(Vec::from_iter(
                             self.recipients
                                 .iter()
                                 .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
@@ -719,7 +721,7 @@ impl From<&ConversationDocument> for Conversation {
         conversation.set_creator(document.creator.clone());
         conversation.set_recipients(document.recipients());
         conversation.set_created(document.created);
-        conversation.set_settings(document.settings);
+        conversation.set_settings(document.settings.clone());
         conversation.set_modified(document.modified);
         conversation.set_favorite(document.favorite);
         conversation.set_description(document.description.clone());
--- a/warp/src/raygun/mod.rs
+++ b/warp/src/raygun/mod.rs
@@ -467,7 +467,7 @@ impl Conversation {
     }
 
     pub fn settings(&self) -> ConversationSettings {
-        self.settings
+        self.settings.clone()
     }
 
     pub fn recipients(&self) -> Vec<DID> {
@@ -525,7 +525,7 @@ impl Conversation {
     }
 }
 
-#[derive(Debug, Hash, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Display)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Display)]
 #[serde(rename_all = "lowercase")]
 #[repr(C)]
 pub enum ConversationSettings {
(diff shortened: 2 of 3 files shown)


warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
+pub struct Role {
+    // Participants assigned to this role
+    participants: Vec<DID>,
+    // Permissions of this role
+    permissions: Vec<Permission>,
+}
+
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+pub enum Permission {
+    All,
```
dariusc93, line 570: Im not all that sure we would want to explicitly have an `All` permission but instead leave it to the creator of the conversation to determine what role should have all the permissions.

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
+    SetRoles,
+    ChangeGroupName,
+    AddParticipantsToGroup,
```
dariusc93, line 573: Might want to split this into two or maybe have separate enumerator to indicate if the permission can add or remove for any explicit type (though when it comes to exporting this to wasm, I can imagine that being a pain unless we change `Permission` to a struct and have a functions to provide the inner indication. Just something to think about)

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
-#[derive(Default, Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Display)]
-#[display(
-    fmt = "Everyone can add participants: {}, Everyone can change name: {}",
-    "if self.members_can_add_participants {\"✅\"} else {\"❌\"}",
-    "if self.members_can_change_name {\"✅\"} else {\"❌\"}"
-)]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Display)]
+#[display(fmt = "Roles: {:?}", "self.roles")]
 #[repr(C)]
 pub struct GroupSettings {
```
dariusc93, line 554: We might want to eventually rename this

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
 #[repr(C)]
 pub struct DirectConversationSettings {}
 
-#[derive(Default, Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Display)]
-#[display(
-    fmt = "Everyone can add participants: {}, Everyone can change name: {}",
-    "if self.members_can_add_participants {\"✅\"} else {\"❌\"}",
-    "if self.members_can_change_name {\"✅\"} else {\"❌\"}"
-)]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Display)]
```
dariusc93, line 551:
```suggestion
#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
-    pub fn set_members_can_add_participants(&mut self, val: bool) {
-        self.members_can_add_participants = val;
+    pub fn set_roles(&mut self, roles: IndexMap<String, Role>) {
+        self.roles = roles;
     }
 
-    pub fn set_members_can_change_name(&mut self, val: bool) {
-        self.members_can_change_name = val;
+    pub fn user_has_permission(&self, id: &DID, permission: Permission) -> bool {
+        for (_, role) in &self.roles {
```
dariusc93, line 590:
```suggestion
        for role in self.roles.values() {
```

dariusc93: Although as I think about it, we could probably reduce this to a single call by using `values` as the iterator and using `any` with the check for permission

warp/src/raygun/mod.rs
```diff
@@ -467,7 +467,7 @@ impl Conversation {
     }
 
     pub fn settings(&self) -> ConversationSettings {
-        self.settings
+        self.settings.clone()
```
dariusc93, line 470: Not required right now but down the road we should try to borrow instead of cloning for the getters here. Would reduce allocations and let downstream handle how they wish to use the data (or if they wish to clone it).

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -341,19 +345,20 @@ impl ConversationDocument {
...
-                            self.recipients
-                                .iter()
-                                .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
-                        )),
+                        (!settings.user_has_permission(&did, Permission::AddParticipantsToGroup))
+                            .then_some(Vec::from_iter(
+                                self.recipients
+                                    .iter()
+                                    .flat_map(|rec| rec.to_string().as_bytes().to_vec()),
+                            )),
```
dariusc93, line 353: I do wonder if we should have this here. I guess for normal group conversations this should be alright but for communities this would be redundant. I guess i would have to think more on that when the type actually comes.

warp/src/raygun/mod.rs
```diff
@@ -525,7 +525,7 @@ impl Conversation {
     }
 }
 
-#[derive(Debug, Hash, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Display)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Display)]
 #[serde(rename_all = "lowercase")]
 #[repr(C)]
 pub enum ConversationSettings {
```
dariusc93, line 531: As we do add permissions in, we could probably go on and get rid of `ConversationSettings`

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
+    roles: IndexMap<String, Role>,
+}
+
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+pub struct Role {
+    // Participants assigned to this role
+    participants: Vec<DID>,
+    // Permissions of this role
+    permissions: Vec<Permission>,
+}
```
dariusc93, line 566: I wonder if we should reduce this to `IndexMap<DID, IndexSet<Permission>>` (or some variant of map/set)? Thoughts on that?

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -128,11 +127,18 @@ impl ConversationDocument {
...
+            ConversationType::Direct
+        } else {
+            ConversationType::Group
         }
     }
+    pub fn has_permission(&self, user: &DID, permission: GroupPermission) -> bool {
+        if let Some(value) = self.permissions.get(&user) {
+            return value.contains(&permission);
+        }
+        return false;
```
dariusc93: We could do something like the following to be more optimized

```suggestion
        self.permissions.iter().any(|(id, perms)| id == user && perms.contains(&permission)) 
```

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -257,14 +259,14 @@ impl ConversationDocument {
 
 impl ConversationDocument {
     pub fn sign(&mut self, keypair: &Keypair) -> Result<(), Error> {
-        if let ConversationSettings::Group(settings) = self.settings {
+        if let ConversationType::Group = &self.conversation_type() {
```
dariusc93, line 262:
```suggestion
        if let ConversationType::Group = self.conversation_type() {
```

warp/src/raygun/mod.rs
```diff
@@ -394,7 +395,7 @@ pub struct Conversation {
     created: DateTime<Utc>,
     favorite: bool,
     modified: DateTime<Utc>,
-    settings: ConversationSettings,
+    permissions: HashMap<DID, Vec<GroupPermission>>,
```
dariusc93, line 398: Do we want to have the permissions here in Conversation or do you feel we should split this into a separate function?

warp/src/raygun/mod.rs
```diff
@@ -103,9 +104,9 @@ pub enum MessageEventKind {
         did_key: DID,
         event: MessageEvent,
     },
-    ConversationSettingsUpdated {
+    ConversationPermissionsUpdated {
         conversation_id: Uuid,
-        settings: ConversationSettings,
+        permissions: HashMap<DID, Vec<GroupPermission>>,
```
dariusc93, line 109: I think it might be best to split this to indicate the identity and what permission was added/removed (maybe a `kind` field). Thoughts?

warp/src/raygun/mod.rs
```diff
@@ -548,41 +548,54 @@ impl Default for ConversationSettings {
...
 pub struct DirectConversationSettings {}
 
-#[derive(Default, Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Display)]
-#[display(
-    fmt = "Everyone can add participants: {}, Everyone can change name: {}",
-    "if self.members_can_add_participants {\"✅\"} else {\"❌\"}",
-    "if self.members_can_change_name {\"✅\"} else {\"❌\"}"
-)]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Display)]
+#[display(fmt = "Roles: {:?}", "self.roles")]
```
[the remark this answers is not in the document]
ashneverdawn: n/a since `ConversationSettings` is now removed

warp/src/raygun/mod.rs
ashneverdawn: thanks. removed.

warp/src/raygun/mod.rs
ashneverdawn: for group permissions, keeping it to a simple enum, since there will only be a few permissions. However, communities will have more complex permissions, so it's worth considering there.

warp/src/raygun/mod.rs
ashneverdawn: removed `ConversationSettings`/`GroupSettings` in favour of `GroupPermissions`

warp/src/raygun/mod.rs
ashneverdawn: removed ConversationSettings

warp/src/raygun/mod.rs
ashneverdawn: removed roles from group chats

warp/src/raygun/mod.rs
ashneverdawn: unchanged in the PR, but I agree 👍

extensions/warp-ipfs/src/store/conversation.rs
ashneverdawn: noted

warp/src/raygun/mod.rs
ashneverdawn: ConversationSettings removed 👍

warp/src/raygun/mod.rs
ashneverdawn: done 👍

extensions/warp-ipfs/examples/messenger.rs
```diff
@@ -514,12 +518,23 @@ async fn main() -> anyhow::Result<()> {
...
+                                Err(e) => {
+                                    writeln!(stdout, "Error getting conversation: {e}")?;
+                                    continue
+                                }
+                                Ok(conversation) => {
+                                    let mut permissions = HashMap::new();
+                                    if open {
+                                        for did in &conversation.recipients() {
+                                            permissions.insert(did.clone(), vec![GroupPermission::AddParticipants, GroupPermission::SetGroupName]);
+                                        }
```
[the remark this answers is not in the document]
ashneverdawn: done

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -51,7 +50,7 @@ pub struct ConversationDocument {
     pub creator: Option<DID>,
     pub created: DateTime<Utc>,
     pub modified: DateTime<Utc>,
-    pub settings: ConversationSettings,
+    pub permissions: HashMap<DID, Vec<GroupPermission>>,
```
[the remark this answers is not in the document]
ashneverdawn: done

extensions/warp-ipfs/src/store/conversation.rs
```diff
@@ -128,11 +127,18 @@ impl ConversationDocument {
     }
 
     pub fn conversation_type(&self) -> ConversationType {
-        match self.settings {
-            ConversationSettings::Direct(_) => ConversationType::Direct,
-            ConversationSettings::Group(_) => ConversationType::Group,
+        if self.recipients.len() <= 2 {
+            ConversationType::Direct
+        } else {
+            ConversationType::Group
```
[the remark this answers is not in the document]
ashneverdawn: I added `ConversationType::from_recipients` method, and call it from `conversation_type` methods on both `ConversationDocument` and `Conversation` to avoid this code duplication.

extensions/warp-ipfs/src/store/conversation.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/conversation.rs
ashneverdawn: done

warp/src/raygun/mod.rs
ashneverdawn: I created a type alias for it `GroupPermissions` as changing the type from `HashMap` to `IndexMap` required a lot of changes. So if we have to change the type again, for example to make it into a struct with more features, we won't have to change it everywhere again.

I also moved `has_permissions` method to the `GroupPermissions` type alias using a trait.

warp/src/raygun/mod.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/conversation.rs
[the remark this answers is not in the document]
ashneverdawn: Maybe this is what you meant we should do all along: I added conversation_type to Conversation and ConversationDocument. This way it can be set on creation. Most of the remaining broken tests were due to incorrect determination of conversation_type. I incorrectly assumed this could be inferred from the number of participants, but there are test cases that create group chats with only 1 or 2 participants.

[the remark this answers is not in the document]
ashneverdawn: Needless to say, I removed `ConversationType::from_recipients` and `conversation_type()` methods now return the value of the field on their respective structs.

dariusc93 (approved):
LGTM. In a separate PR we can update the trait member "update_conversation_permissions" to add a conversion to it for the permissions itself.

PR #606, opened 2024-09-27
Add Communities feature

ashneverdawn:
**What this PR does** 📖

- Adds the ability to create communities which can contain multiple messaging channels, roles, and permissions.


warp/src/raygun/community.rs
```diff
@@ -0,0 +1,224 @@
...
+    ViewChannel,
+    EditName,
+    EditDescription,
+    SendMessages,
+    DeleteMessages,
+}
+
+#[async_trait::async_trait]
+pub trait RayGunCommunity: Sync + Send {
+    async fn create_community(&mut self, _owner: &DID, _name: &str) -> Result<Community, Error> {
```
dariusc93, line 120: we should remove the owner parameter since the caller of the function would be the creator.

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+    pub fn permissions(&self) -> &CommunityChannelPermissions {
+        &self.permissions
+    }
+}
+#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum CommunityChannelType {
+    Text,
+    Voice,
+}
```
dariusc93, line 97: Not sure if we should have "Text" and "Voice" as a type since `RayGun` would be mostly for chat. We could probably let the frontend impl handle that in some way or maybe change `Voice` to something else that represents a non-text chat.

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,224 @@
...
+    async fn edit_community_permissions(
+        &mut self,
+        _community_id: Uuid,
+        _permissions: CommunityPermissions,
+    ) -> Result<(), Error> {
+        Err(Error::Unimplemented)
+    }
+    async fn add_community_member(&mut self, _community_id: Uuid, _member: DID) -> Result<(), Error> {
+        Err(Error::Unimplemented)
+    }
```
dariusc93, line 175: I wonder if this should have a option to treat it as an invite of sorts so the user would have the option to join or ignore. Thoughts?

warp/src/raygun/community.rs
ashneverdawn: done

warp/src/raygun/community.rs
ashneverdawn: From warp's side, Text and Voice channels are really both just Text channels and behave exactly the same. (Voice channels still need a chat in them) We just need a way to store the 'intention' the channel was created for in Warp. That way the ui can know that a particular channel is intended to be voice enabled and also to know to render the chat differently. (Like a Discord Voice channel's embedded text chat)

Renamed these to `Standard` and `VoiceEnabled` to make it a little more clear.

warp/src/raygun/mod.rs
```diff
@@ -29,6 +31,8 @@ pub enum RayGunEventKind {
     ConversationArchived { conversation_id: Uuid },
     ConversationUnarchived { conversation_id: Uuid },
     ConversationDeleted { conversation_id: Uuid },
+
+    CommunityCreated { community_id: Uuid },
```
dariusc93, line 35: Thoughts about adding a flag to `ConversationCreated` to indicate the type of conversation it is?

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,416 @@
...
+    name: String,
+    members: IndexSet<DID>,
+}
+impl CommunityRole {
+    pub fn id(&self) -> RoleId {
+        self.id
+    }
+    pub fn name(&self) -> &String {
+        &self.name
+    }
```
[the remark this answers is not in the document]
ashneverdawn: done

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,416 @@
...
+    created: DateTime<Utc>,
+    expiry: Option<DateTime<Utc>>,
+}
+impl CommunityInvite {
+    pub fn id(&self) -> Uuid {
+        self.id
+    }
+    pub fn target_user(&self) -> &Option<DID> {
+        &self.target_user
+    }
```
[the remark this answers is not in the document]
ashneverdawn: done

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,416 @@
...
+impl Community {
+    pub fn id(&self) -> Uuid {
+        self.id
+    }
+    pub fn name(&self) -> &String {
+        &self.name
+    }
+    pub fn description(&self) -> &Option<String> {
+        &self.description
+    }
```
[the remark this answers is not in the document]
ashneverdawn: done

warp/src/raygun/community.rs
```diff
@@ -0,0 +1,416 @@
...
+impl CommunityChannel {
+    pub fn id(&self) -> Uuid {
+        self.id
+    }
+    pub fn name(&self) -> &String {
+        &self.name
+    }
+    pub fn description(&self) -> &Option<String> {
+        &self.description
+    }
```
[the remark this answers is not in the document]
ashneverdawn: done

extensions/warp-ipfs/src/store/community.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl CommunityDocument {
+    pub fn id(&self) -> Uuid {
+        self.id
+    }
+
+    pub fn name(&self) -> &String {
+        &self.name
+    }
```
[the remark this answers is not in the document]
ashneverdawn: done

warp/src/raygun/mod.rs
ashneverdawn: Yeah it could come in handy if in the ui we need screens where we show only DirectConversations or only GroupConversations. That way we wouldn't need to fetch a conversation if we weren't interested in that ConversationType. (At the moment though, they are both listed together on the same list, like discord does, so they all need to be fetched anyway.) Alternatively it could be split into different events, which would achieve the same effect, perhaps even better since the ui would just directly subscribe to the events they need instead of checking the conversation type on each event emitted. Either way, probably best to do that in a separate PR from this one.

ashneverdawn:
Closing in favor of #617 where I resolved merge issues

PR #612, opened 2024-10-14
refactor: spawn task per conversation

### LINKED ISSUE #601: raygun: process each conversation in its own task
Currently, each conversation has a stream that it would process in its own task, but would send to the data received to the current task to be processed as the stream is polled in the current task. This allows for conversations to be process within the task in FIFO manner. While this works great as it is now, it may slow things down it comes to expanding to large subset of conversations or if there is. The best solution here would be to have each conversation in its own respected task as it was before and process them specifically for that conversation, updating the document accordingly. 

Note:
- When performing actions (sending, replying, editing, etc), we should send it to the respected task, in queue, to be handled concurrently, with the events the user send taking priority over processing anything from stream
- When the conversation is deleted, the task itself should be terminated after the messages are deleted.
- each task should store its own respected queue, or at least a reference of what needs to be sent out.

dariusc93:
**What this PR does** 📖
- move conversation into its own task
- use channels to communicate with task
- remove redundant enums out of stream

- resolves #601

**Additional comments** 🎤
- In a separate PR will have a implementation for a controller, which would have a common interface set for creating, managing and destroying conversations (relates to communities, though this may allow us to split direct and group conversations into their own documents and thus own tasks as well). 
- A future change would allow us to queue commands as well as the tasks to allow `ConversationTask` to be made into a future or stream (likely a future as we would still send a response via oneshot).
- There is a slight performance increase by splitting conversations off into their own task
  - there may be more of an increase as reduce some of the async calls and process internal task as we poll the future.


ashneverdawn (approved):
Thanks! Nice job splitting up that massive message.rs file. Looking forward to future PRs building on top of this. 👍

PR #627, opened 2024-11-13
chore: fix tests for ci

ashneverdawn:
**What this PR does** 📖

Makes community tests wait for `RayGunEventKind::CommunityInvite` event before accepting invites. This should solve the flaky tests that sometimes fail in the ci.

### DIFF
--- a/extensions/warp-ipfs/tests/community.rs
+++ b/extensions/warp-ipfs/tests/community.rs
@@ -1,14 +1,14 @@
 mod common;
 #[cfg(test)]
 mod test {
-    use futures::StreamExt;
+    use futures::{stream::FuturesUnordered, FutureExt, Stream, StreamExt, TryStreamExt};
     use std::time::Duration;
     use warp::raygun::{
         community::{
             Community, CommunityChannelPermission, CommunityChannelType, CommunityInvite,
             CommunityPermission, RayGunCommunity,
         },
-        MessageEventKind, MessageEventStream, RayGunEventKind, RayGunEventStream, RayGunStream,
+        MessageEventKind, MessageEventStream, RayGunEventKind, RayGunStream,
     };
 
     #[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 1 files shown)


extensions/warp-ipfs/tests/community.rs
```diff
@@ -51,6 +51,19 @@ mod test {
...
+    ) -> Result<RayGunEventKind, std::io::Error> {
+        crate::common::timeout(timeout, async {
+            loop {
+                if let Some(event) = stream.next().await {
+                    return event;
+                }
+            }
+        })
+        .await
+    }
```
dariusc93, line 66: we could probably simplify this by doing


```rust
    async fn next_event<S: Stream + Unpin>(
        stream: &mut S,
        timeout: Duration,
    ) -> Result<S::Item, std::io::Error> {
        crate::common::timeout(timeout, async {
            loop {
                if let Some(event) = stream.next().await {
                    return event;
                }
            }
        })
        .await
    }
 ```
 
 That way we dont need separate functions for each stream

Comment (dariusc93) [reply extensions/warp-ipfs/tests/community.rs:66]: We could probably ditch the loop as well since it would be expected that the stream wont terminate so we could also remove the and expect the next event. Maybe doing 

```rust
    async fn next_event<S: Stream + Unpin>(
        stream: &mut S,
        timeout: Duration,
    ) -> Result<S::Item, std::io::Error> {
        crate::common::timeout(
            timeout,
            stream
                .next()
                .map(|item| item.expect("stream not to terminate")),
        )
        .await
    }
```

dariusc93 (commented):
LGTM. Left a suggestion to simplify things a bit

extensions/warp-ipfs/tests/community.rs
ashneverdawn: done 👍

extensions/warp-ipfs/tests/community.rs
```diff
@@ -24,21 +24,21 @@ mod test {
...
     ) -> Result<(), std::io::Error> {
         let mut events = vec![];
         for stream in streams {
-            events.push(next_event(stream, timeout.clone()).await?);
+            events.push(next_msg_event(stream, timeout.clone()).await?);
         }
         for event in events {
             assert_eq!(event, expected_event);
         }
         Ok(())
```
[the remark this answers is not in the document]
ashneverdawn: with the first one some of the tests were failing. Didn't look too much into it and just used your second suggestion which worked fine 👍

dariusc93 (approved):

PR #632, opened 2024-11-19
feat: String based permission nodes

Flemmli97:
**What this PR does** 📖

Rewrites the community permission system to use string based permissions instead of enums. String based permission have the form `some.permission` and have the advantage that they easily expandable and also allows hierarchy checking.

A permission `some.permission` will check under the hood if the user has the node
`some.permission` and `some` with the deeper one taking priority. If `some.permission` is defined it will use that permission. Otherwise it will check `some`.

The builtin permissions are still in an enum to allow ease of usage. The API to check a permission accepts anything that is representable as a String.

### DIFF
--- a/tools/enum_macro/src/lib.rs
+++ b/tools/enum_macro/src/lib.rs
@@ -1,0 +1,116 @@
+extern crate proc_macro2;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput};
+
+///
+/// Implements Display for enum values with formatting and snake_case conversion
+/// This is used for CommunityPermissions
+/// E.g.
+/// ```rust
+/// #[derive(EnumFormatting)]
+/// pub enum SomeEnum {
+///     #[permission(formatting="prefix.{}")]
+///     SomeValue1,
+///     SomeValue2
+/// }
+///
+/// assert_eq!(SomeEnum::SomeValue1.to_string(), "prefix.some_value_1")
+/// assert_eq!(SomeEnum::SomeValue2.to_string(), "some_value_2")
+/// ```
+#[proc_macro_derive(EnumFormatting, attributes(permission))]
+pub fn permission_node(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+    if let Data::Enum(enum_data) = &input.data {
+        let mut variants = vec![];
+        for variant in &enum_data.variants {
+            let mut format = None;
+            if let Some(attr) = variant
+                .attrs
+                .iter()
+                .find(|attr| attr.path().is_ident("permission"))
+            {
+                if let syn::Meta::List(l) = &attr.meta {
+                    if let Err(err) = l.parse_nested_meta(|meta| {
+                        if meta.path.is_ident("formatting") {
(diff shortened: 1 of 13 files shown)


tools/enum_macro/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+                pub fn values() -> &'static[#name] {
+                    &[ #(#name::#variants),* ]
+                }
+            }
+        };
+        TokenStream::from(expanded)
+    } else {
+        panic!("Only Enums are supported");
+    }
+}
```
dariusc93, line 116: Wouldnt it be better to just use serde to rename the enum than to use a macro?

dariusc93 (commented):
Could you explain how this has benefits over just using enums, given that the permissions themselves are fixed, having them as strings altogether seems a bit redundant. Internally when is serialized to a string, it can represent as this node-style and it could be easily deserialize to an enum.

tools/enum_macro/src/lib.rs
Flemmli97: serde can only to something like upper/lowercase etc. they dont have anything like formatting sadly

Flemmli97:
> Could you explain how this has benefits over just using enums, given that the permissions themselves are fixed, having them as strings altogether seems a bit redundant. Internally when is serialized to a string, it can represent as this node-style and it could be easily deserialize to an enum.

initially it was so you could set (and check) arbitrary permission as otherwise they would be hardcoded. It would allow other devs etc. to make use of this system too and defining their own permissions. though i agree that the way its currently written still doesnt allow that

tools/enum_macro/src/lib.rs
dariusc93: you can use `#[serde(rename="some.permission.node")]` for each enum variant. Ie

```rust
#[derive(Serialize, Deserialize)]
enum PermissionNode {
    #[serde(rename="add.identity")]
    AddIdentity,
    #[serde(rename="remove.identity")]
    RemoveIdentity
}
```

Flemmli97: yes but with that its a bit decoupled from the actual enum name which i didnt like

dariusc93: Thats only an example though. It could be named anything, however serde would handle that though when it comes to deserializing it back to the enum.

Flemmli97: i meant that serde does not give access to the enum name inside renaming. so if you e.g. rename the enum itself `AddIdentity` you also need to update the serialized name. its a bit nitpicky but initially i wanted the enums purely for sanity checks within warp and allow arbitrary strings as permissions

dariusc93: In that case you would want to serialize it manually if you dont want to update the rename attr macro, although this shouldnt exactly be a problem to rename it either (though to take caution against breaking changes too).

Flemmli97: i can change it back to serde if you think its better. i personally think manually defining the string everytime is a bit cumbersome

Flemmli97:
hmm not sure why running `cargo clippy -- -D warnings` does not throw these linter errors...

dariusc93:
Test failing due to invalid field (should be `permissions`) and invalid type.

PR #635, opened 2024-11-25
chore: update config.rs

Tinrii:
changed max upload file size value from 100MB to 20GB as it should be

**What this PR does** 📖
fixes what dev **ashneverdawn** wrote as note that i stated in this pull request
https://github.com/Satellite-im/UplinkWeb/pull/881
https://github.com/Satellite-im/UplinkWeb/pull/881

**Additional comments** 🎤

### DIFF
--- a/extensions/warp-ipfs/src/lib.rs
+++ b/extensions/warp-ipfs/src/lib.rs
@@ -2066,7 +2066,7 @@ impl RayGunCommunity for WarpIpfs {
         &mut self,
         community_id: Uuid,
         channel_id: Uuid,
-        message_id: Option<Uuid>,
+        message_id: Uuid,
     ) -> Result<(), Error> {
         self.messaging_store()?
             .delete_community_channel_message(community_id, channel_id, message_id)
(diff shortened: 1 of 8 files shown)


dariusc93:
Hey! Thanks for the PR. 

I did leave a comment on https://github.com/Satellite-im/UplinkWeb/pull/881#issuecomment-2498942898 but in all I dont mind the change, but do keep in mind that it is set to 100MB as a balance between desktop/native (including mobile) and web/wasm. Additionally, it leaves room for additional files to be added before it hits the max storage limit. Also, we would not be able to set it beyond 2GB due to wasm32 address space target limit (see comment for more details). So if we are to set this value, it should be set to 2GB.

Tinrii:
hello, i have changed it again and set it to 2GB instead of 20GB 😄

ashneverdawn (commented):
I suggest following the pattern used by `max_storage_size`:
```Some(2 * 1024 * 1024 * 1024)```

Tinrii:
updated 😄

PR #640, opened 2024-12-03
Feat: community message attachments

ashneverdawn:
**What this PR does** 📖

Implements message attachments for community messages

### DIFF
--- a/extensions/warp-ipfs/src/store/message/community_task.rs
+++ b/extensions/warp-ipfs/src/store/message/community_task.rs
@@ -29,9 +29,7 @@ use warp::raygun::community::{
     CommunityPermission, CommunityRole, RoleId,
 };
 use warp::raygun::{
-    AttachmentEventStream, ConversationImage, Location, MessageEvent, MessageOptions,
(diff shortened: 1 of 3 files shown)


extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -3304,29 +3306,125 @@ impl CommunityTask {
...
+        channel_id: Uuid,
+        message_id: Option<Uuid>,
+        locations: Vec<Location>,
+        messages: Vec<String>,
     ) -> Result<(Uuid, AttachmentEventStream), Error> {
-        Err(Error::Unimplemented)
+        let own_did = &self.identity.did_key();
+        if !self.document.has_channel_permission(
+            own_did,
+            &CommunityChannelPermission::SendMessages,
```
dariusc93, line 3317: I wonder if we should split this so it has two permission checks, one for sending messages and another for attaching a file. Thoughts?

dariusc93 (approved):

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: Sounds good. Done 👍

PR #649, opened 2024-12-18
feat: community invites that anyone can use

ashneverdawn:
**What this PR does** 📖

- implements the ability to create invites with no target which anyone can use.
- instead calling `accept_community_invite` which required both the `community_id` and `invite_id`, now a user calls `request_join_community` which only requires `community_id`. An event is then received (`CommunityJoinRejected`/`CommunityJoined`) based on whether the join was successful or not. To succeed, an invite with valid criteria must exist.
- creating a targeted invite used to immediately add the user to the community. They still had to accept the invite, but that was just to enable them to perform any actions. Now the user isn't part of the community until they've actually joined.


extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
 
         Ok(())
     }
+    async fn process_join_event(&mut self, msg: Message) -> Result<(), Error> {
+        let data = PayloadMessage::<Vec<u8>>::from_bytes(&msg.data)?;
```
dariusc93, line 1287: If we are not going to encrypt the events, we could just have it be `PayloadMessage<CommunityJoinEvents>` instead and skip `serde_json::from_slice`. Thoughts?

dariusc93: Just as a note, depending on how the message is sent, we could encrypt the message.

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+        let event = serde_json::from_slice::<CommunityJoinEvents>(data.message()).map_err(|e| {
+            tracing::warn!(community_id = %community_id, sender = %data.sender(), error = %e, "Failed to deserialize message");
+            e
+        })?;
+        match event {
+            CommunityJoinEvents::Join => {
+                let mut is_invited = false;
+                for (_, invite) in &self.document.invites {
+                    if let Some(expiry) = invite.expiry {
+                        if expiry < Utc::now() {
```
dariusc93, line 1299: I wonder if we should move `Utc::now` into its own variable outside of the loop and use that when checking to see if it expired that way if there is a large subset of invites, it doesnt expire while it is processing, unless that is the intent. If so, then we probably should probably have a function apart of the invite that checks to see if it expired.

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+        let community_id = self.community_id;
+        let sender = data.sender().to_did()?;
+        let event = serde_json::from_slice::<CommunityJoinEvents>(data.message()).map_err(|e| {
+            tracing::warn!(community_id = %community_id, sender = %data.sender(), error = %e, "Failed to deserialize message");
+            e
+        })?;
+        match event {
+            CommunityJoinEvents::Join => {
+                let mut is_invited = false;
+                for (_, invite) in &self.document.invites {
```
dariusc93, line 1297: I wonder if we could optimize this by using an iterator.

extensions/warp-ipfs/src/store/message.rs
```diff
@@ -1028,20 +1036,39 @@ impl MessageStore {
...
-                response: tx,
-            })
-            .await;
-        rx.await.map_err(anyhow::Error::from)?
+        match inner.community_task.get(&community_id) {
+            None => {
+                let keypair = inner.root.keypair();
+
+                let event = CommunityJoinEvents::DeleteInvite { invite_id };
+                let bytes = serde_json::to_vec(&event)?;
```
dariusc93, line 1044: Do we need to convert this to a vector instead of just supplying `event`. to `PayloadBuilder::new`?

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+                for (id, invite) in &self.document.invites {
+                    if let Some(target) = &invite.target_user {
+                        if &sender == target {
+                            invites_to_remove.push(id.clone());
+                        }
+                    }
+                }
+                for id in &invites_to_remove {
+                    self.document.invites.swap_remove(id);
+                }
```
dariusc93, line 1335: We could optimize this with an iterator or even to go as far as to just remove the items within the initial for loop instead of pushing it into a vector.

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+
+                self.set_document().await?;
+
+                self.send_single_community_event(
+                    &sender.clone(),
```
dariusc93, line 1340: Shouldnt need to borrow a clone. 

```suggestion
                    &sender,
```

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+                    .is_some_and(|target| target == sender)
+                {
+                    return Err(Error::InvalidCommunityInvite);
+                }
+
+                self.document.invites.swap_remove(&invite_id);
+                self.set_document().await?;
+
+                self.send_single_community_event(
+                    &sender.clone(),
```
dariusc93, line 1374:
```suggestion
                    &sender,
```

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+                        if &sender != target {
+                            continue;
+                        }
+                    }
+                    is_invited = true;
+                    break;
+                }
+                if !is_invited {
+                    self.send_single_community_event(
+                        &sender.clone(),
```
dariusc93, line 1313:
```suggestion
                        &sender,
```

extensions/warp-ipfs/src/store/message/community_task.rs
```diff
@@ -1275,6 +1283,105 @@ impl CommunityTask {
...
+                        community_id: self.community_id,
+                        community_document: Some(self.document.clone()),
+                    },
+                )
+                .await?;
+
+                if !self.discovery.contains(&sender).await {
+                    let _ = self.discovery.insert(&sender).await;
+                }
+                if let Err(_e) = self.request_key(&sender.clone()).await {}
```
dariusc93, line 1351:
```suggestion
                if let Err(_e) = self.request_key(&sender).await {}
```

ashneverdawn: done

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: done.

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: done

extensions/warp-ipfs/src/store/message.rs
ashneverdawn: removed. thanks

extensions/warp-ipfs/src/store/message/community_task.rs
ashneverdawn: Skipped `serde_json::from_slice`, thanks. As discussed, we can leave the join event plaintext here since in the Payload refactor you are working on rn the message will be encrypted there.

dariusc93 (approved):

